PR #1, opened 2020-03-12
Requires admin to use the tools and temporarily hides SSO code for later

staubina:
fix


src/index.jsx
```diff
@@ -23,13 +23,18 @@ function supportLinks() {
       <h3>Support Tools</h3>
       <ul>
         <li><Link to="/users">Search Users</Link></li>
+        <li>TEST: </li>
```
davidjoy: Nix

davidjoy (approved):

src/index.jsx
staubina: 👍

PR #9, opened 2020-03-30
General cleanup.

davidjoy:
See each commit for a breakdown of the changes in here - they're probably easier to review that way.

The first commit is letting the linter auto-fix a bunch of stuff, so it's a bit verbose.  The others are all smaller.


src/users/api.js
```diff
@@ -144,15 +150,18 @@ export async function patchEntitlement({
     return data;
   } catch (error) {
     if (error.customAttributes.httpErrorStatus === 400) {
+      // We don't have good error handling in the app for any errors that may have come back
+      // from the API, so we log them to the console and tell the user to go look.  We would
+      // never do this in a customer-facing app.
```
staubina, line 156: I will add to the list of tasks to improve this messaging.  Good catch.

staubina (approved):

staubina:
I would be interested in learning from you about the following if you have time over coffee at some point.

* User messaging updates

staubina:
Thank you @davidjoy!  Should we squash these commits before merging?

davidjoy:
Yeah, I'll give it a squash.  And happy to talk about User Messages whenever. :)

PR #14, opened 2020-04-03
fix: fixes a search issue with invalid characters

staubina:


crice100 (approved):
I approved so you can merge if you need to, but IMHO it's always good to add unit tests with regex changes.

fysheets (approved):

PR #18, opened 2020-04-17
Fix table sorting bug

bbaker6225:
This provides a fix for the issue where some columns of the table were
not being sorted correctly. The issue is that some of the columns have
objects, so they would never be sorted differently when keyed into.
This adds a utility method for getting the internal value of an element
in the table, which lets us appropriately sort things like the
courseId.
Additionally changes to displaying N/A for missing dates in the table.

CR-2058

UI after:
image

### DIFF
--- a/src/users/Enrollments.jsx
+++ b/src/users/Enrollments.jsx
@@ -30,11 +30,13 @@ export default function Enrollments({
       courseId: <a href={`${getConfig().LMS_BASE_URL}/courses/${result.courseId}`} rel="noopener noreferrer" target="_blank">{result.courseId}</a>,
       courseStart: result.courseStart,
       courseEnd: result.courseEnd,
-      upgradeDeadline: result.verifiedUpgradeDeadline,
+      // Set upgrade deadline to null if it doesn't exist to work around https://github.com/moment/moment/issues/3180
+      upgradeDeadline: result.verifiedUpgradeDeadline || null,
       created: result.created,
       reason: result.manualEnrollment ? result.manualEnrollment.reason : '',
       lastModifiedBy: result.manualEnrollment ? result.manualEnrollment.enrolledBy : '',
-      lastModified: result.manualEnrollment ? result.manualEnrollment.timeStamp : '',
+      // Set last modified to null if the timestamp doesn't exist to work around https://github.com/moment/moment/issues/3180
+      lastModified: result.manualEnrollment ? result.manualEnrollment.timeStamp || null : '',
       active: result.isActive ? 'True' : 'False',
       mode: result.mode,
       actions: (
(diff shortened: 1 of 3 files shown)


bbaker6225:
@davidjoy I looked into the `value` vs. `displayValue` method a bit and it was turning out to be pretty invasive. I'm happy to try again if you think there is an issue with this approach, or if we need to do something else for handling the date weirdness

src/users/Enrollments.jsx
```diff
@@ -30,11 +30,13 @@ export default function Enrollments({
       courseId: <a href={`${getConfig().LMS_BASE_URL}/courses/${result.courseId}`} rel="noopener noreferrer" target="_blank">{result.courseId}</a>,
       courseStart: result.courseStart,
       courseEnd: result.courseEnd,
-      upgradeDeadline: result.verifiedUpgradeDeadline,
+      // Set upgrade deadline to null if it doesn't exist to work around https://github.com/moment/moment/issues/3180
+      upgradeDeadline: result.verifiedUpgradeDeadline || null,
```
bbaker6225: This and `lastModified` may go back to the way they were, depending on the result of https://edx-internal.slack.com/archives/CTVD85CUV/p1587127044248100

src/users/sort.js
```diff
@@ -1,10 +1,27 @@
+/**
+ * Given a keyed value from an element on the table, returns its internal value for sorting if it exists.
+ * This lets us properly support course ids that are embedded in links, as well as other objects.
+ */
+function getInternalValue(keyedValue) {
+  if (keyedValue) {
+    const { props = {} } = keyedValue;
+    return props.children;
```
davidjoy, line 8: I think we should go ahead with it for now, but it’s probably brittle long-term for us to assume props.children is going to be a sortable value…

We could add a warning here if children is not something we deem sortable - i.e., not a primitive value, basically.  If it's an object or function, it'd save future-developers from running into basically the same problem if we make a fuss so they see something's wrong.

src/users/sort.js
```diff
@@ -1,10 +1,27 @@
...
 export default function sort(firstElement, secondElement, key, direction) {
   const directionIsAsc = direction === 'asc';
 
-  if (firstElement[key] > secondElement[key]) {
+  const firstKeyedValue = firstElement[key];
+  const firstInternalValue = getInternalValue(firstKeyedValue);
+  const secondKeyedValue = secondElement[key];
+  const secondInternalValue = getInternalValue(secondKeyedValue);
+
+  if (firstKeyedValue > secondKeyedValue || firstInternalValue > secondInternalValue) {
```
davidjoy, line 21: These `if` statements on 21 and 24 feel a bit funny to me, in that we compare the "Keyed" values and rely on them returning `false` before we even look at the internal values.  I'm not sure it's possible, but if for some reason they're comparable/return true, we won't even check that the internal values exist, which are sorta "higher priority" if they exist.

Might be that if our internal values are non-undefined, we just compare them, otherwise we fall back to the Keyed values.

davidjoy (approved):
A few suggestions, but 👍 in spirit.

bbaker6225:
Closing in favor of #20

PR #19, opened 2020-04-17
Copying header-edx in and turning it into a support header.

davidjoy:
Screen Shot 2020-04-17 at 11 19 05 AM

Also ran an `npm audit fix`


bbaker6225 (approved):
Looks great locally. I haven't reviewed the code too closely yet, but I don't think that needs to block things unless you'd like a closer look

Dillon-Dumesnil (approved):
I wouldn't necessarily trust my reviews on frontend matters, but nothing looks wrong to me!

davidjoy:
FWIW - most of the new code is snagged directly from frontend-component-header-edx, so it's been in production for quite some time.  I mostly just tweaked Header.jsx and the way the submenuContext's render.

PR #20, opened 2020-04-17
Fix table sorting bug

bbaker6225:
Fixes the table sorting bug by associating each cell datum with a
sortable internal value. This also adds a display value to each cell,
which is the priority for rendering that data in the table.
As part of the new rendering logic for table data, this also adds a
utility function for formatting dates which cleans things up a bit.


bbaker6225:
image

davidjoy (approved):

PR #21, opened 2020-04-17
Adding ability to use email in URL

Dillon-Dumesnil:
We already support searching by email or username,
but the URL only supported username. This allows the URL
to use either username or email.

This PR also has various refactors to make the code more DRY

### DIFF
--- a/src/users/UserPage.jsx
+++ b/src/users/UserPage.jsx
@@ -3,7 +3,7 @@ import React, {
 } from 'react';
 import PropTypes from 'prop-types';
 
-import { camelCaseObject, history } from '@edx/frontend-platform';
+import { camelCaseObject, getConfig, history } from '@edx/frontend-platform';
 import { Link } from 'react-router-dom';
 import PageLoading from '../PageLoading';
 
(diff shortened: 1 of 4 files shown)


src/users/UserPage.jsx
```diff
@@ -26,12 +30,21 @@ export default function UserPage({ match }) {
...
+    if (userIdentifier) {
+      if (isEmail(userIdentifier)) {
+        history.replace(`/users/?email=${userIdentifier}`);
+      } else if (isValidUsername(userIdentifier)) {
+        history.replace(`/users/?username=${userIdentifier}`);
+      }
     }
-    if (result.errors.length > 0) {
-      result.errors.forEach(error => add(error));
+    console.log(errors);
```
davidjoy: Can probably remove this console log if we're adding them to the alert list.

Dillon-Dumesnil: That would have been my debugging and I thought I removed it. My bad. One sec

davidjoy:
This seems nice!  

One thing I _think_ you might be able to do is to use the `username` in `data`, rather than even having the username useState hook.  Whenever we search, `data` gets set and should include it, and we should be able to then just pass it down to the sub-components that need it.

Might be worth 5 minutes to see if that's possible - if it's hairy, then I think it's fine to leave.

Cool!

davidjoy (approved):
See one comment about possibly being able to get rid of `username` state.

staubina (approved):
I am not seeing anything blocking here.  I recommend we all test on Stage though.

Dillon-Dumesnil:
> This seems nice!
> 
> One thing I _think_ you might be able to do is to use the `username` in `data`, rather than even having the username useState hook. Whenever we search, `data` gets set and should include it, and we should be able to then just pass it down to the sub-components that need it.
> 
> Might be worth 5 minutes to see if that's possible - if it's hairy, then I think it's fine to leave.
> 
> Cool!

Wonderful callout. You are absolutely correct. Removed the set username logic in favor of data.user.username!

staubina (approved):

PR #22, opened 2020-04-17
WIP: Add 'Create new enrollment' button

bbaker6225:
This adds in most of the frontend code needed to support the 'Create
new enrollment' button/action. An important thing to note is that the
API that we plan to use: /support/enrollment/{user} does not currently
support creating new enrollments, only updating them. To get this work
across the line, we'll have to update that API in the LMS.

CR-2068

image

### DIFF
--- a/src/users/api.js
+++ b/src/users/api.js
@@ -304,7 +304,44 @@ export async function postEnrollmentChange({
         {
           code: null,
           dismissible: true,
-          text: 'There was an error submitting this entitlement. Check the JavaScript console for detailed errors.',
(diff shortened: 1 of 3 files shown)


src/users/api.js
```diff
@@ -304,7 +304,44 @@ export async function postEnrollmentChange({
         {
           code: null,
           dismissible: true,
-          text: 'There was an error submitting this entitlement. Check the JavaScript console for detailed errors.',
+          text: 'There was an error changing this enrollment. Check the JavaScript console for detailed errors.',
```
bbaker6225, line 307: Just a typo fix

PR #23, opened 2020-04-17
Improves history management by using history.push

davidjoy:
We use history.push when adding a username or email, but generally use replace for errors and going back to /users

Also updating the document title with the search term.

### DIFF
--- a/src/users/UserPage.jsx
+++ b/src/users/UserPage.jsx
@@ -37,17 +37,25 @@ export default function UserPage({ location }) {
     return !!(searchValue && searchValue.match(USERNAME_REGEX));
   }
 
-  function processSearchResult(searchValue, result) {
-    if (isEmail(searchValue)) {
-      history.replace(`/users/?email=${searchValue}`);
-    } else if (isValidUsername(searchValue)) {
-      history.replace(`/users/?username=${searchValue}`);
+  function pushHistoryIfChanged(nextUrl) {
+    if (nextUrl !== location.pathname + location.search) {
+      history.push(nextUrl);
     }
+  }
 
+  function processSearchResult(searchValue, result) {
     if (result.errors.length > 0) {
       result.errors.forEach(error => add(error));
-      history.replace('/users/');
+      history.replace('/users');
+      document.title = 'Support Tools | edX';
+    } else if (isEmail(searchValue)) {
+      pushHistoryIfChanged(`/users/?email=${searchValue}`);
+      document.title = `Support Tools | edX | ${searchValue}`;
+    } else if (isValidUsername(searchValue)) {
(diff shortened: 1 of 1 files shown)


src/users/UserPage.jsx
```diff
@@ -62,7 +70,7 @@ export default function UserPage({ location }) {
         type: 'error',
         topic: 'general',
       });
-      history.replace('/users/');
+      history.replace('/users');
```
Dillon-Dumesnil, line 73: Yeah, I noticed that when testing. Thanks for catching!

src/users/UserPage.jsx
```diff
@@ -37,17 +37,25 @@ export default function UserPage({ location }) {
...
     }
+  }
 
+  function processSearchResult(searchValue, result) {
     if (result.errors.length > 0) {
       result.errors.forEach(error => add(error));
-      history.replace('/users/');
+      history.replace('/users');
+      document.title = 'Support Tools | edX';
+    } else if (isEmail(searchValue)) {
```
Dillon-Dumesnil, line 51: Nice reorder

Dillon-Dumesnil (approved):
LGTM

staubina (approved):

src/users/UserPage.jsx
```diff
@@ -37,17 +37,25 @@ export default function UserPage({ location }) {
     return !!(searchValue && searchValue.match(USERNAME_REGEX));
   }
 
-  function processSearchResult(searchValue, result) {
-    if (isEmail(searchValue)) {
-      history.replace(`/users/?email=${searchValue}`);
-    } else if (isValidUsername(searchValue)) {
-      history.replace(`/users/?username=${searchValue}`);
+  function pushHistoryIfChanged(nextUrl) {
+    if (nextUrl !== location.pathname + location.search) {
```
davidjoy, line 41: I discovered that it creates a history entry even if the value is the same.

src/users/UserPage.jsx
```diff
@@ -37,17 +37,25 @@ export default function UserPage({ location }) {
...
+      history.push(nextUrl);
     }
+  }
 
+  function processSearchResult(searchValue, result) {
     if (result.errors.length > 0) {
       result.errors.forEach(error => add(error));
-      history.replace('/users/');
+      history.replace('/users');
```
davidjoy, line 49: In the case of errors, we just use replace.  That's not meaningful history.

src/users/UserPage.jsx
```diff
@@ -114,13 +122,23 @@ export default function UserPage({ location }) {
     }
   }, [userIdentifier]);
 
+  useEffect(() => {
```
davidjoy, line 125: This useEffect is necessary to get the page to re-search when the user uses the forward/back buttons.

PR #26, opened 2020-05-11
Fix issue with CourseSummary table

staubina:
Changes to the Table to support sorting made it so that an Object
cannot be sent to the Component as Table data. For now I am switching to
a simpler table to just show the data.

### DIFF
--- a/src/Table.jsx
+++ b/src/Table.jsx
@@ -140,7 +140,7 @@ export default class Table extends React.Component {
   }
 
   /*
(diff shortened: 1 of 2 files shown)


bbaker6225 (approved):
I think the best "somewhat quick" approach would be to eventually update so that all the easy sortable fields are handled by their value like on the user summary page, and then we come up with a more elegant sorting strategy for the arrays of course runs later.

However, I think this is fine for a super quick fix once tests pass

PR #29, opened 2020-07-23
Fix enrollment tool.

waheedahmed:
If course has only one mode (e.g. professional) and enrollment mode is different from it (e.g. audit), there is no way to trigger onChange leading to submitting the old mode again instead of new mode.

PROD-1548


adzuci (approved):

waheedahmed:
Thank you @adzuci!

PR #30, opened 2020-07-27
Add support for SSO endpoint and display data on user page

bseverino:
[MST-305](https://openedx.atlassian.net/jira/software/projects/MST/boards/584)

Display SSO records in the "Search Users" tool. Includes a modal to display extra data if necessary.

Requires this PR to be merged: https://github.com/edx/edx-platform/pull/24614

---

Screen Shot 2020-07-28 at 11 33 00 AM

Screen Shot 2020-07-28 at 11 33 15 AM

### DIFF
--- a/src/users/UserSummary.jsx
+++ b/src/users/UserSummary.jsx
@@ -1,10 +1,19 @@
-import React from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
+import { Modal, Button } from '@edx/paragon';
 
 import Table from '../Table';
 import formatDate from '../dates/formatDate';
 
-export default function UserSummary({ userData, verificationData }) {
+export default function UserSummary({
+  userData,
+  verificationData,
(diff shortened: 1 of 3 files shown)


src/users/UserSummary.jsx
```diff
@@ -57,9 +56,24 @@ export default function UserSummary({ userData, verificationData }) {
       dataName: 'Verified',
       dataValue: verificationData.isVerified ? 'Yes' : 'No',
     },
+  ];
 
+  const ssoColumns = [
+    {
+      label: 'Provider',
```
schenedx, line 72: I think it would be create to also display `created` and `modified` dates. Also, it would be great if you can show a link while hiding the `extra_data` field values. When user clicks on that link, the data in the `extra_data` field would be shown.

schenedx (changes requested):
See my inline comments.

schenedx (approved):
👍

PR #31, opened 2020-07-29
Give support the ability to disable/enable user accounts

bseverino:
[MST-307: Add manage user tool to MFE](https://openedx.atlassian.net/browse/MST-307)

@edx/masters-devs-cosmonauts 

In the "Search Users" tool, shows the password status for the user ("Usable" or "Unusable") and adds a button to toggle this status. Also includes some styling fixes to fit this information.

---

Screen Shot 2020-07-29 at 12 00 51 PM

### DIFF
--- a/src/users/UserSummary.jsx
+++ b/src/users/UserSummary.jsx
@@ -2,6 +2,7 @@ import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import { Modal, Button } from '@edx/paragon';
 
+import { postTogglePasswordStatus } from './api';
 import Table from '../Table';
 import formatDate from '../dates/formatDate';
 
@@ -9,10 +10,21 @@ export default function UserSummary({
   userData,
   verificationData,
   ssoRecords,
+  changeHandler,
 }) {
   const [modalIsOpen, setModalIsOpen] = useState(false);
   const [extraDataTitle, setExtraDataTitle] = useState('');
   const [ssoExtraData, setSsoExtraData] = useState([]);
+
+  const PASSWORD_STATUS = {
+    USABLE: 'Usable',
+    UNUSABLE: 'Unusable',
+  };
+
+  const togglePasswordStatus = () => {
+    postTogglePasswordStatus(userData.username);
+    changeHandler();
+  };
 
   const userAccountData = [
     {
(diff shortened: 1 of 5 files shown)


src/users/api.js
```diff
@@ -282,3 +290,10 @@ export async function postEnrollmentChange({
     };
   }
 }
+
+export async function postTogglePasswordStatus(user) {
+  const { data } = await getAuthenticatedHttpClient().post(
+    `${getConfig().LMS_BASE_URL}/support/manage_user/${user}`,
```
schenedx, line 296: This is an existing API on edx-platform, correct?

schenedx (commented):
Just a question. I think you should get @staubina's review on this change as well.

src/users/api.js
bseverino: @schenedx Yes, I didn't have to add anything to edx-platform for this one

src/users/UserSummary.jsx
```diff
@@ -114,29 +120,42 @@ export default function UserSummary({
...
           <h4>Account</h4>
           <Table
             data={userAccountData}
             columns={columns}
           />
+          <Button
+            className={`${userData.passwordStatus === 'Usable' ? 'btn-outline-danger' : 'btn-outline-primary'} toggle-password`}
+            onClick={togglePasswordStatus}
+          >
+            {userData.passwordStatus === 'Usable' ? 'Disable User' : 'Enable User'}
```
staubina: Should this be a constant? 'Usable' is not clear to me.

staubina (commented):
Overall this PR looks good to me. Please make sure the Support team is aware this is coming though in the #support-tools channel.

src/users/UserSummary.jsx
bseverino: The existing endpoint either returns a string 'Usable' or 'Unusable' based on the status; a boolean might make more sense, but I'm not sure if that should be addressed with this PR.

schenedx (approved):
👍

src/users/UserSummary.jsx
staubina: The constant would make it more clear to me what we are comparing here. Whenever I see a string being used multiple times I tend to think it should probably be a named constant in case that changes. Also at times, the name of the constant is more readable than the actual string.

staubina: At the end of the day, this is more of a "nit" on code style than it is a functional issue. I am going to approve and leave the decision to change up to you.

staubina (approved):

src/users/UserSummary.jsx
bseverino: That makes sense, thank you for the feedback! I will update this PR and merge.

PR #32, opened 2020-07-31
Hide disable/enable user button

bseverino:
Hides this feature until it can be better implemented with feedback from support.


schenedx (approved):
👍

PR #34, opened 2020-08-26
MST-348 Add the verification data detail modal so support can see those info

schenedx:
@staubina and @edx/masters-devs-cosmonauts Please review
Add a button and a model to the ID Verification panel to show list of ID verification for the user
See screenshot:

Screen Shot 2020-08-26 at 3 31 48 PM

Screen Shot 2020-08-26 at 3 32 45 PM

### DIFF
--- a/src/users/UserSummary.jsx
+++ b/src/users/UserSummary.jsx
@@ -12,9 +12,12 @@ export default function UserSummary({
   ssoRecords,
   changeHandler,
 }) {
-  const [modalIsOpen, setModalIsOpen] = useState(false);
-  const [extraDataTitle, setExtraDataTitle] = useState('');
+  const [ssoModalIsOpen, setSsoModalIsOpen] = useState(false);
+  const [idvModalIsOpen, setIdvModalIsOpen] = useState(false);
+  const [extraSsoDataTitle, setSsoExtraDataTitle] = useState('');
+  const [detailIdvDataTitle, setDetailIdvDataTitle] = useState('');
   const [ssoExtraData, setSsoExtraData] = useState([]);
+  const [detailIdvData, setDetailIdvData] = useState([]);
   const userToggleVisible = false;
   // TO-DO: Only expose "Disable/Enable User" for specific roles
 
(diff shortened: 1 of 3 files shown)


src/users/UserSummary.jsx
```diff
@@ -123,7 +149,31 @@ export default function UserSummary({
...
+          Show
+        </Button>
+      ) : 'N/A',
+      value: result.extraData,
+    },
+  }));
+
+  // Modal to display extra data for Idv records
+  const openIDVModal = (title, data) => {
+    setDetailIdvData(data);
```
staubina: The expiration date time that is in the modal is not formatted into a readable format. It will help keep the UI consistent if we run the date format on it. This might be easily solved by just moving where/when the date is formatted. 

It looks like you already format the date for the summary block.

staubina (changes requested):
Overall I am really excited to see this data added and I think support will be also. I am not excited about the format of the data in the Modal. Could we maybe find a way to add a little padding around the table or something to make it a little easier to read? Or make it so that the modal is wider?

staubina:
FYI @kashifch, I wanted you to know this was an upcoming change.

src/users/UserSummary.jsx
```diff
@@ -123,7 +149,31 @@ export default function UserSummary({
       displayValue: Object.keys(result.extraData).length > 0 ? (
         <Button
           className="btn-link px-0"
-          onClick={() => openModal(result.provider, result.extraData)}
+          onClick={() => openSSOModal(result.provider, result.extraData)}
+        >
+          Show
```
alangsto, line 154: NIT: This is really small, but the padding around this element looks different, making it seem like "Show" isn't in line with the other columns.

alangsto (approved):
This looks great, non-blocking comment.

schenedx:
@staubina and @alangsto Addressed your feedbacks.
The modal dialog being too small is not a problem I am going to solve in this PR. The Modal is a paragon component. In order to make the modal bigger, we need to bump up the version of Paragon. That breaks all the buttons and collapsible components on the page. I have a fix there but that's just too complicated for this PR. 
I'll find a way to solve that later.

staubina (approved):

PR #42, opened 2020-09-21
Updating Renovate.json

awaisdar001:
This renovate configuration adds the following:

- only renovate on Monday morning at 12 am.
- automerge minor and patch versions if they pass tests
- automerge major, minor, and patch versions of linters and testers, if they pass tests
- Turns out the default is UTC so set the timezone to `America/New_York` 
- Make no updates to branches when not scheduled

I think the automerge still needs some repo permissions set up to be able to work right.

## Schedule Config Issue: 
Adding `schedule` as defined in the docs.  => https://docs.renovatebot.com/faq/

### Screenshot 1: 
Screen Shot 2020-09-21 at 2 55 18 PM

## rebaseStalePrs Config Issue: 
Also, there is a setting int the file, `"rebaseStalePrs": true,` I am not sure if this is applied because in the docs they recommended using `rebaseWhen` settings. 
https://docs.renovatebot.com/configuration-options/

### Screenshot 2:
Screen Shot 2020-09-21 at 2 54 32 PM

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,10 +1,12 @@
 {
   "extends": [
     "config:base",
-    "schedule:earlyMondays"
+    ":automergeLinters",
+    ":automergeTesters",
+    ":automergeMinor",
+    ":noUnscheduledUpdates"
   ],
-  "patch": {
-    "automerge": true
-  },
-  "rebaseStalePrs": true
+  "rebaseStalePrs": true,
+  "schedule": ["after 12am and before 08:30am on Monday"],
+  "timezone": "America/New_York"
 }


renovate.json
```diff
@@ -6,5 +6,7 @@
   "patch": {
```
awaisdar001, line 6: There is a setting on L#4 ` "schedule:earlyMondays"` but I don't think that matches with the docs. 

Screen Shot 2020-09-21 at 2 55 18 PM

renovate.json
```diff
@@ -6,5 +6,7 @@
...
     "automerge": true
   },
-  "rebaseStalePrs": true
+  "rebaseStalePrs": true,
+  "timezone": "America/Los_Angeles",
```
awaisdar001: Turns out the default is UTC.

staubina: If we are setting a non-UTC time should we use East coast time? Just curious why we are choosing LA time?

awaisdar001: I was following the docs and found thought that a USA based time would make sense. If you think we should change that to EST I can do that too.

renovate.json
```diff
@@ -1,10 +1,12 @@
...
+    ":automergeTesters",
+    ":automergeMinor",
+    ":noUnscheduledUpdates"
   ],
-  "patch": {
-    "automerge": true
-  },
-  "rebaseStalePrs": true
+  "rebaseStalePrs": true,
+  "schedule": ["at 12:00 am on Monday"],
```
albemarle: I don't believe this will work, because it is just one point in time; you will need to give it a time window.

renovate.json
```diff
@@ -1,10 +1,12 @@
...
+  "timezone": "America/Los_Angeles"
```
albemarle: Probably going to make people happier to use America/New_York .  :-)

renovate.json
awaisdar001: Thanks, adding `after 12am and before 08:30am on Monday`

awaisdar001:
@albemarle thank you for the review. I appreciate that. 

I have updated the PR with the provided suggestions. Please take a look when you can. Thanks.

albemarle (approved):

PR #62, opened 2020-10-12
Recursively fetch data from entitlement paginated API

awaisdar001:
Recursively fetch data from entitlement paginated API
-

The data from the Entitlement API is paginated with 50 records per page. This is the case when _support-tools-frontend_ has to fetch all the records from the API when the user has more than 50 entitlements. Therefore, This PR iteratively fetch all the pages which contain the data we need using the recursive function (A recursive function is a function that calls itself).

[PROD-2170](https://openedx.atlassian.net/browse/PROD-2170)
FYI -- @staubina

### DIFF
--- a/src/users/api.js
+++ b/src/users/api.js
@@ -4,10 +4,15 @@ import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
 const EMAIL_REGEX = '^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\\.[a-zA-Z]{2,}$';
 const USERNAME_REGEX = '^[\\w.@_+-]+$';
 
-export async function getEntitlements(username) {
-  const { data } = await getAuthenticatedHttpClient().get(
-    `${getConfig().LMS_BASE_URL}/api/entitlements/v1/entitlements/?user=${username}`,
-  );
+export async function getEntitlements(username, page = 1) {
+  const baseURL = `${getConfig().LMS_BASE_URL}/api/entitlements/v1/entitlements/`;
+  const queryString = `user=${username}&page=${page}`;
+  const { data } = await getAuthenticatedHttpClient().get(`${baseURL}?${queryString}`);
+  if (data.next !== null) {
+    const nextPageData = await getEntitlements(username, data.current_page + 1);
+    data.results = data.results.concat(nextPageData.results);
+    return data;
+  }
   return data;
 }
 
@@ -146,7 +151,6 @@ export async function getAllUserData(userIdentifier) {
   let enrollments = [];
   let verificationStatus = null;
   let ssoRecords = null;
-
   try {
     user = await getUser(userIdentifier);
   } catch (error) {


src/users/api.js
```diff
@@ -4,10 +4,15 @@ import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
...
-    `${getConfig().LMS_BASE_URL}/api/entitlements/v1/entitlements/?user=${username}`,
-  );
+export async function getEntitlements(username, page = 1) {
+  const baseURL = `${getConfig().LMS_BASE_URL}/api/entitlements/v1/entitlements/`;
+  const queryString = `user=${username}&page=${page}`;
+  const { data } = await getAuthenticatedHttpClient().get(`${baseURL}?${queryString}`);
+  if (data.next !== null) {
+    const nextPageData = await getEntitlements(username, data.current_page + 1);
+    data.results = data.results.concat(nextPageData.results);
+    return data;
```
DawoudSheraz, line 14: Why recursive? This can be handled via a loop. Plus, each call is maintaining its own stack too.

src/users/api.js
```diff
@@ -4,10 +4,15 @@ import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
 const EMAIL_REGEX = '^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\\.[a-zA-Z]{2,}$';
 const USERNAME_REGEX = '^[\\w.@_+-]+$';
 
-export async function getEntitlements(username) {
-  const { data } = await getAuthenticatedHttpClient().get(
-    `${getConfig().LMS_BASE_URL}/api/entitlements/v1/entitlements/?user=${username}`,
-  );
+export async function getEntitlements(username, page = 1) {
+  const baseURL = `${getConfig().LMS_BASE_URL}/api/entitlements/v1/entitlements/`;
```
DawoudSheraz, line 8: should this be moved to some central config file? Now that code is being updated, it will be good to get some nits done alongside too.

awaisdar001: Yes, that can be done. +1

src/users/api.js
awaisdar001: This version of the code is structurally much simpler, without the need for any variables to maintain the current iteration state. With a recursive function, this same information is stored in the stack. I found this much more concise.

```
export async function getEntitlements(username, page = 1) {
  const baseURL = `${getConfig().LMS_BASE_URL}/api/entitlements/v1/entitlements/`;
  aggregateData = [];
  while (true) {
    const queryString = `user=${username}&page=${page}`;
    const { data } = await getAuthenticatedHttpClient().get(`${baseURL}?${queryString}`);
    aggregateData = aggregateData.concat(data.results);
    if (data.nextPage) {
      page = data.nextPage;
    } else {
      break;
    }
  }
  return aggregateData;
}
```

src/users/api.js
awaisdar001: We have already defined `getConfig().LMS_BASE_URL` in frontend-platfrom, what would you recommend? 
Adding variable in .env files e.g.
`API_URL_ENTITLEMENTS=/api/entitlements/v1/entitlements/`

and have a helper function 
```
export const getEntitlementsAPIBaseURL = () => `${getConfig().LMS_BASE_URL}${env.API_URL_ENTITLEMENTS}`;
```

What do you think?

src/users/api.js
DawoudSheraz: We get the total number of pages in the paginated response, so no need for a while loop.

src/users/api.js
DawoudSheraz: yes, this would be better.

src/users/api.js
```diff
@@ -4,10 +4,15 @@ import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
...
+  const queryString = `user=${username}&page=${page}`;
```
asadazam93, line 9: Only the page changes here. We can keep the username as part of the baseURL?

src/users/api.js
asadazam93: ++

DawoudSheraz (approved):

src/users/api.js
awaisdar001: `username` is part of the request query string which would be concatenated with `baseURL` as `?`. `username` is part of the baseurl.

asadazam93 (approved):
looks good

PR #63, opened 2020-10-12
Added disable user history/confirmation

asadazam93:
### Description
Added Disable user button with a confirmation pop up that allow you to add a comment/reason. Also added user enable/disable history

UI Images
Screenshot 2020-10-15 at 3 37 00 PM
Screenshot 2020-10-15 at 3 36 49 PM

**Sandbox**
N/A

### Reviewers
If you've been tagged for review, please check your corresponding box once you've given the :+1:.
- [ ] @awaisdar001 
- [ ] @DawoudSheraz 

### Post-review
- [ ] Rebase and squash commits

### DIFF
--- a/src/users/UserSummary.jsx
+++ b/src/users/UserSummary.jsx
@@ -1,6 +1,6 @@
 import React, { useState } from 'react';
 import PropTypes from 'prop-types';
-import { Modal, Button } from '@edx/paragon';
+import { Modal, Button, Input } from '@edx/paragon';
 
 import { postTogglePasswordStatus } from './api';
 import Table from '../Table';
@@ -14,11 +14,15 @@ export default function UserSummary({
 }) {
   const [ssoModalIsOpen, setSsoModalIsOpen] = useState(false);
   const [idvModalIsOpen, setIdvModalIsOpen] = useState(false);
+  const [disableUserModalIsOpen, setDisableUserModalIsOpen] = useState(false);
+  const [disableHistoryModalIsOpen, setDisableHistoryModalIsOpen] = useState(false);
+  const [comment, setComment] = useState('');
+  const [userPasswordHistoryData, setUserPasswordHistoryData] = useState([]);
   const [extraSsoDataTitle, setSsoExtraDataTitle] = useState('');
   const [detailIdvDataTitle, setDetailIdvDataTitle] = useState('');
   const [ssoExtraData, setSsoExtraData] = useState([]);
   const [detailIdvData, setDetailIdvData] = useState([]);
-  const userToggleVisible = false;
+  const userToggleVisible = true;
   // TO-DO: Only expose "Disable/Enable User" for specific roles
 
   const PASSWORD_STATUS = {
@@ -27,7 +31,7 @@ export default function UserSummary({
   };
 
   const togglePasswordStatus = () => {
-    postTogglePasswordStatus(userData.username);
+    postTogglePasswordStatus(userData.username, comment);
     changeHandler();
   };
 
@@ -58,7 +62,7 @@ export default function UserSummary({
     },
     {
       dataName: 'Password Status',
-      dataValue: userData.passwordStatus,
+      dataValue: userData.passwordStatus.status,
     },
   ];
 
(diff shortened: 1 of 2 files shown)


src/users/UserSummary.jsx
```diff
@@ -14,11 +14,15 @@ export default function UserSummary({
 }) {
   const [ssoModalIsOpen, setSsoModalIsOpen] = useState(false);
   const [idvModalIsOpen, setIdvModalIsOpen] = useState(false);
+  const [disableUserModalIsOpen, setDisableUserModalIsOpen] = useState(false);
+  const [disableHistoryModalIsOpen, setDisableHistoryModalIsOpen] = useState(false);
```
DawoudSheraz, line 18: although the previous code is written is adding IsOpen at the end, in my opinion, the var name should be like IsDisableUserModelOpen. This gives the intent of being a boolean right from the start.

src/users/api.js
```diff
@@ -161,7 +161,7 @@ export async function getAllUserData(userIdentifier) {
     enrollments = await getEnrollments(user.username);
     verificationStatus = await getUserVerificationStatus(user.username);
     ssoRecords = await getSsoRecords(user.username);
-    user.passwordStatus = await getUserPasswordStatus(user.username);
+    user = await getUserPasswordStatus(user.username);
```
DawoudSheraz: userPasswordStatus is more logical.

src/users/UserSummary.jsx
```diff
@@ -249,6 +286,42 @@ export default function UserSummary({
...
+          open={disableUserModalIsOpen}
+          buttons={[
+            <Button
+              variant="success"
+              className="btn-outline-primary"
+              onClick={togglePasswordStatus}
+            >
+              Confirm
+            </Button>,
+          ]}
```
DawoudSheraz: Why two modals?

DawoudSheraz (commented):
Please add how the UI is looking with button and modal.

src/users/api.js
asadazam93: I did not change the variable name. Previously it was only fetching the passwordStatus attribute from the api response but I needed all of the data

src/users/UserSummary.jsx
asadazam93: Both modals are different.

src/users/UserSummary.jsx
```diff
@@ -249,6 +291,42 @@ export default function UserSummary({
             />
           )}
         />
+        <Modal
+          open={disableHistoryModalIsOpen}
+          onClose={() => setDisableHistoryModalIsOpen(false)}
+          title="Enable/Disable History"
+          body={(
+            <Table
+              data={disableHistoryData}
```
awaisdar001: `defaultSortedColumn` and `defaultSortDirection`

src/users/UserSummary.jsx
```diff
@@ -249,6 +291,42 @@ export default function UserSummary({
...
+              Confirm
+            </Button>,
+          ]}
+          onClose={() => setDisableUserModalIsOpen(false)}
+          title={`${userData.status === PASSWORD_STATUS.USABLE ? 'Disable user confirmation' : 'Enable user confirmation'}`}
+          body={(
+            <div>
+              <label htmlFor="userIdentifier">Reason: </label>
+              <Input
+                name="userIdentifier"
```
awaisdar001: userIdentifier`

src/users/UserSummary.jsx
```diff
@@ -262,7 +340,8 @@ UserSummary.propTypes = {
     isActive: PropTypes.bool,
     country: PropTypes.string,
     dateJoined: PropTypes.string,
-    passwordStatus: PropTypes.string,
+    status: PropTypes.string,
+    disableHistory: PropTypes.shape([]),
```
awaisdar001: `disableHistory`

src/users/api.js
awaisdar001: ++

src/users/UserSummary.jsx
asadazam93: A user probably will not have a long history of disabling/enabling so I guess this would be an overhead

DawoudSheraz (approved):

awaisdar001 (approved):

PR #79, opened 2020-12-18
add special chars in email regex check

DawoudSheraz:
### Description
Email regex does not include special characters that are allowed in the emails. The PR adds a few of them, taking reference from [django email validator](https://github.com/django/django/blob/83fbaa92311dd96e330496a0e443ea71b9c183e2/django/core/validators.py#L157).

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -1,4 +1,4 @@
-const EMAIL_REGEX = '^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\\.[a-zA-Z]{2,}$';
+const EMAIL_REGEX = '^[a-zA-Z0-9\'!#$&*._%+-]+@[a-zA-Z0-9.-]+\\.[a-zA-Z]{2,}$';
 const USERNAME_REGEX = '^[\\w.@_+-]+$';
 
 // todo: No need for !!. This operation is not affecting the value.
--- a/src/utils/index.test.js
+++ b/src/utils/index.test.js
@@ -1,7 +1,7 @@
 import { isEmail, isValidUsername } from './index';
 
 describe('Test Utils', () => {
-  const validEmails = ['staff@email.com', 'admin@email.co.og'];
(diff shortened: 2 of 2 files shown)


awaisdar001 (approved):

awaisdar001:
Just curious, do we allow special char emails in the platform?

DawoudSheraz:
> Just curious, do we allow special char emails in the platform?

I tested the account creation on local using email with some characters and the email was valid. The backend is using django form's EmailField to validate email. EmailField uses a very complex regex which I have shared in the description. I have only added a few of the special characters from that regex.

src/utils/index.js
```diff
@@ -1,4 +1,4 @@
-const EMAIL_REGEX = '^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\\.[a-zA-Z]{2,}$';
+const EMAIL_REGEX = '^[a-zA-Z0-9\'!#$&*._%+-]+@[a-zA-Z0-9.-]+\\.[a-zA-Z]{2,}$';
```
azanbinzahid, line 1: https://github.com/edx/edx-platform/blob/master/lms/envs/common.py#L1094

Isn't this the allowed username regex in edX-platform?

DawoudSheraz: Yeah, but the change is for email only

azanbinzahid, line 1: `&` needs to be added with `\&`

azanbinzahid (changes requested):

src/utils/index.js
DawoudSheraz: No, it will work without escaping. Only ' needed to be escape due to str format. The identifiers are between []. You can verify via any online JS regex tester(https://regexr.com/).

azanbinzahid: Oh yes, correct. they have different meaning outside [] but not within.

azanbinzahid (approved):

PR #82, opened 2021-01-20
add last login field in user summary component

DawoudSheraz:
### Description
Adds last login time for user summary component. The field was added in user_api in https://github.com/edx/edx-platform/pull/26087

Screenshot 2021-01-20 at 3 27 14 PM
Screenshot 2021-01-20 at 3 27 33 PM


saadyousafarbi (approved):
LGTM

PR #83, opened 2021-01-21
Add ability to expire Entitlements

Ali-D-Akbar:
## Add ability to expire entitlements in support tool 
[PROD-2132](https://openedx.atlassian.net/browse/PROD-2132)

An `Expire` button is added in the action column of Entitlements as shown in the image:

Screenshot 2021-01-21 at 6 49 58 PM

A form will be shown upon clicking the `Expire` button. The user can add a comment about the reason to expire the entitlement.
Screenshot 2021-01-21 at 6 51 00 PM

Submitting this will send a patch request to api to set `expiredAt` field. Note that the button will be disabled once the entitlement is expired. 
Screenshot 2021-01-21 at 6 52 52 PM

### DIFF
--- a/src/users/data/api.js
+++ b/src/users/data/api.js
@@ -221,24 +221,11 @@ export async function getCourseData(courseUUID) {
 }
 
 export async function patchEntitlement({
-  uuid,
-  action,
-  unenrolledRun = null,
-  comments = null,
+  uuid, requestData,
 }) {
   try {
     const { data } = await getAuthenticatedHttpClient().patch(
-      `${getConfig().LMS_BASE_URL}/api/entitlements/v1/entitlements/${uuid}/`,
-      {
-        expired_at: null,
-        support_details: [
-          {
-            unenrolled_run: unenrolledRun,
-            action,
-            comments,
-          },
-        ],
-      },
+      `${getConfig().LMS_BASE_URL}/api/entitlements/v1/entitlements/${uuid}/`, requestData,
     );
     return data;
   } catch (error) {
@@ -265,27 +252,11 @@ export async function patchEntitlement({
 }
 
 export async function postEntitlement({
-  user,
-  courseUuid,
-  mode,
-  action,
-  comments = null,
+  requestData,
 }) {
   try {
     const { data } = await getAuthenticatedHttpClient().post(
-      `${getConfig().LMS_BASE_URL}/api/entitlements/v1/entitlements/`,
-      {
-        course_uuid: courseUuid,
-        user,
-        mode,
-        refund_locked: true,
-        support_details: [
-          {
-            action,
-            comments,
-          },
-        ],
-      },
+      `${getConfig().LMS_BASE_URL}/api/entitlements/v1/entitlements/`, requestData,
     );
     return data;
   } catch (error) {
--- a/src/users/entitlements/CreateEntitlementForm.jsx
+++ b/src/users/entitlements/CreateEntitlementForm.jsx
@@ -5,15 +5,13 @@ import {
 } from '@edx/paragon';
 import classNames from 'classnames';
 
-import UserMessagesContext from '../user-messages/UserMessagesContext';
-import AlertList from '../user-messages/AlertList';
-import { postEntitlement, patchEntitlement } from './data/api';
+import UserMessagesContext from '../../user-messages/UserMessagesContext';
+import AlertList from '../../user-messages/AlertList';
+import { postEntitlement } from '../data/api';
+import { CREATE } from './EntitlementActions';
+import { EntitlementPropTypes, EntitlementDefaultProps } from './PropTypes';
 
-export const REISSUE = 'reissue';
-export const CREATE = 'create';
-
-export default function EntitlementForm({
-  formType,
+export default function CreateEntitlementForm({
   entitlement,
   changeHandler,
   closeHandler,
@@ -27,50 +25,36 @@ export default function EntitlementForm({
 
   const submit = useCallback(() => {
     clear('entitlements');
-    if (formType === CREATE) {
-      postEntitlement({
+    postEntitlement({
+      requestData: {
+        course_uuid: courseUuid,
         user,
-        courseUuid,
         mode,
-        action: CREATE,
-        comments,
-      }).then((result) => {
-        if (result.errors !== undefined) {
-          result.errors.forEach(error => add(error));
-        } else {
-          changeHandler();
-        }
-      });
-    } else if (formType === REISSUE) {
-      patchEntitlement({
-        uuid: entitlement.uuid,
-        action: REISSUE,
-        unenrolledRun: entitlement.enrollmentCourseRun,
-        comments,
-      }).then((result) => {
-        if (result.errors !== undefined) {
-          result.errors.forEach(error => add(error));
-        } else {
-          changeHandler();
-        }
-      });
-    }
+        refund_locked: true,
+        support_details: [{
+          action: CREATE,
+          comments,
+        }],
+      },
+    }).then((result) => {
+      if (result.errors !== undefined) {
+        result.errors.forEach(error => add(error));
+      } else {
+        changeHandler();
+      }
+    });
   });
-
-  const isReissue = formType === REISSUE;
-  const title = isReissue ? 'Re-issue Entitlement' : 'Create Entitlement';
 
   return (
     <section className="card mb-3">
       <form className="card-body">
         <AlertList topic="entitlements" className="mb-3" />
-        <h4 className="card-title">{title}</h4>
+        <h4 className="card-title">Create Entitlement</h4>
         <h5 className="card-subtitle">All fields are required</h5>
         <div className="form-group">
           <label htmlFor="courseUuid">Course UUID</label>
           <Input
             type="text"
-            disabled={isReissue}
             id="courseUuid"
             name="courseUuid"
             value={courseUuid}
@@ -81,7 +65,6 @@ export default function EntitlementForm({
           <label htmlFor="mode">Mode</label>
           <Input
             type="select"
-            disabled={isReissue}
             id="mode"
             name="mode"
             defaultValue={mode}
@@ -128,42 +111,15 @@ export default function EntitlementForm({
   );
 }
 
-EntitlementForm.propTypes = {
-  formType: PropTypes.string.isRequired,
-  entitlement: PropTypes.shape({
-    uuid: PropTypes.string.isRequired,
-    courseUuid: PropTypes.string.isRequired,
-    enrollmentCourseRun: PropTypes.string,
-    created: PropTypes.string.isRequired,
-    modified: PropTypes.string.isRequired,
-    expiredAt: PropTypes.string,
-    mode: PropTypes.string.isRequired,
-    orderNumber: PropTypes.string,
-    supportDetails: PropTypes.arrayOf(PropTypes.shape({
-      supportUser: PropTypes.string,
-      action: PropTypes.string,
-      comments: PropTypes.string,
-      unenrolledRun: PropTypes.string,
-    })),
-    user: PropTypes.string.isRequired,
-  }),
+CreateEntitlementForm.propTypes = {
+  entitlement: EntitlementPropTypes,
   user: PropTypes.string.isRequired,
   changeHandler: PropTypes.func.isRequired,
   closeHandler: PropTypes.func.isRequired,
   forwardedRef: PropTypes.shape({ current: PropTypes.instanceOf(Element) }),
 };
 
-EntitlementForm.defaultProps = {
-  entitlement: {
-    uuid: '',
-    courseUuid: '',
-    created: '',
-    modified: '',
-    expiredAt: '',
-    mode: 'verified',
-    orderNumber: '',
-    supportDetails: [],
-    user: '',
-  },
+CreateEntitlementForm.defaultProps = {
+  entitlement: EntitlementDefaultProps,
   forwardedRef: null,
 };
--- a/src/users/entitlements/EntitlementActions.jsx
+++ b/src/users/entitlements/EntitlementActions.jsx
@@ -1,0 +1,3 @@
+export const REISSUE = 'reissue';
+export const EXPIRE = 'expire';
+export const CREATE = 'create';
--- a/src/users/entitlements/EntitlementForm.jsx
+++ b/src/users/entitlements/EntitlementForm.jsx
@@ -1,0 +1,65 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { CREATE, REISSUE, EXPIRE } from './EntitlementActions';
+import CreateEntitlementForm from './CreateEntitlementForm';
+import ReissueEntitlementForm from './ReissueEntitlementForm';
+import ExpireEntitlementForm from './ExpireEntitlementForm';
+import { EntitlementPropTypes, EntitlementDefaultProps } from './PropTypes';
+
+export default function EntitlementForm({
+  formType,
+  entitlement,
+  changeHandler,
+  closeHandler,
+  user,
+  forwardedRef,
+}) {
+  if (formType === CREATE) {
+    return (
+      <CreateEntitlementForm
+        key="entitlement-create-form"
+        user={user}
+        entitlement={entitlement}
+        changeHandler={changeHandler}
+        closeHandler={closeHandler}
+        forwardedRef={forwardedRef}
+      />
+    );
+  } if (formType === EXPIRE) {
+    return (
+      <ExpireEntitlementForm
+        key="entitlement-expire-form"
+        user={user}
+        entitlement={entitlement}
+        changeHandler={changeHandler}
+        closeHandler={closeHandler}
+        forwardedRef={forwardedRef}
+      />
+    );
+  } if (formType === REISSUE) {
+    return (
+      <ReissueEntitlementForm
+        key="entitlement-reissue-form"
+        user={user}
+        entitlement={entitlement}
+        changeHandler={changeHandler}
+        closeHandler={closeHandler}
+        forwardedRef={forwardedRef}
+      />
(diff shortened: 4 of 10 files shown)


saadyousafarbi (commented):
Before review had some questions regarding the AC of the ticket being shared:

- **Record the name of the support person expiring the entitlement, the time of expiration, and the reason for expiration.**
Is this being captured anywhere on this new flow?


- **If there are two entitlements appearing in a course for a user, expiring one shouldn’t impact the other in any way.**
Have you tested this scenario, with learners having two entitlements for the same course? Looks like it shouldnt be a problem... But just to make sure!

Ali-D-Akbar:
> * **Record the name of the support person expiring the entitlement, the time of expiration, and the reason for expiration.**
>   Is this being captured anywhere on this new flow?


The API handles this very purpose as you can see below.
Screenshot 2021-01-22 at 3 20 01 PM

> * **If there are two entitlements appearing in a course for a user, expiring one shouldn’t impact the other in any way.**
>   Have you tested this scenario, with learners having two entitlements for the same course? Looks like it shouldnt be a problem... But just to make sure!



We're making a patch request to the API with the **uuid of the entitlement** instead of `course_uuid`. This will only expire the specific entitlement.

src/users/data/api.js
```diff
@@ -227,17 +228,16 @@ export async function patchEntitlement({
...
       {
-        expired_at: null,
-        support_details: [
-          {
-            unenrolled_run: unenrolledRun,
-            action,
-            comments,
-          },
-        ],
+        expired_at: ((action === EXPIRE) ? now : null),
```
awaisdar001: This API layer shouldn't construct data -- and it should use the data. so, please if you are passing this request data it's better (for clarity) to pass from the calling function. 

In the future if we want to reuse `patchEntitlement` method we would not need to come back here and add another `if` condition.

src/users/entitlements/EntitlementActions.jsx
```diff
@@ -0,0 +1,3 @@
+export const REISSUE = 'reissue';
```
awaisdar001, line 1: 👍

src/users/entitlements/EntitlementCreateForm.jsx
```diff
@@ -5,16 +5,13 @@ import {
...
+import { postEntitlement } from '../data/api';
+import { CREATE } from './EntitlementActions';
 
-export const REISSUE = 'reissue';
-export const CREATE = 'create';
 
-
-export default function EntitlementForm({
-  formType,
+export default function EntitlementCreateForm({
```
awaisdar001: Better => `CreateEntitlementForm`

src/users/entitlements/EntitlementCreateForm.jsx
```diff
@@ -28,38 +25,22 @@ export default function EntitlementForm({
...
+        result.errors.forEach(error => add(error));
+      } else {
+        changeHandler();
+      }
+    });
   });
 
-  const isReissue = formType === REISSUE;
-  const title = isReissue ? 'Re-issue Entitlement' : 'Create Entitlement';
+  let title = 'Create Entitlement';
```
awaisdar001: why `let` than `const` here?

src/users/entitlements/EntitlementExpireForm.jsx
```diff
@@ -0,0 +1,142 @@
...
+  const [courseUuid, setCourseUuid] = useState(entitlement.courseUuid);
+  const [mode, setMode] = useState(entitlement.mode);
+  const [comments, setComments] = useState('');
+  const { add, clear } = useContext(UserMessagesContext);
+
+  const submit = useCallback(() => {
+    clear('entitlements');
+    patchEntitlement({
+      uuid: entitlement.uuid,
+      action: EXPIRE,
```
awaisdar001: Here you should pass the data you want to patch, don't expose app logic to the API layer.

src/users/entitlements/EntitlementExpireForm.jsx
```diff
@@ -0,0 +1,142 @@
...
+    }).then((result) => {
+      if (result.errors !== undefined) {
+        result.errors.forEach(error => add(error));
+      } else {
+        changeHandler();
+      }
+    });
+  });
+
+  let title = 'Expire Entitlement';
```
awaisdar001: declarations should go on top.

src/users/entitlements/EntitlementExpireForm.jsx
```diff
@@ -0,0 +1,142 @@
...
+            name="courseUuid"
+            value={courseUuid}
+            disabled={true}
+            onChange={(event) => setCourseUuid(event.target.value)}
+          />
+        </div>
+        <div className="form-group">
+          <label htmlFor="mode">Mode</label>
+          <Input
+            type="select"
```
awaisdar001: why `type=select` in readonly mode? why can't it be `text` instead?

src/users/entitlements/EntitlementForm.jsx
```diff
@@ -0,0 +1,108 @@
...
+  formType,
+  entitlement,
+  changeHandler,
+  closeHandler,
+  user,
+  forwardedRef,
+}) {
+  const [courseUuid, setCourseUuid] = useState(entitlement.courseUuid);
+  const [mode, setMode] = useState(entitlement.mode);
+  const [comments, setComments] = useState('');
```
awaisdar001: `comments` => `comment`

src/users/entitlements/EntitlementForm.jsx
```diff
@@ -0,0 +1,108 @@
...
+        <EntitlementCreateForm
+          key="entitlement-create-form"
+          user={user}
+          entitlement={entitlement}
+          changeHandler={changeHandler}
+          closeHandler={closeHandler}
+          forwardedRef={forwardedRef}
+        />
+      );
+    default:
```
awaisdar001: Let's use `if/else` here.  switch/case is preferred if the number of cases are more than 5 AFAIK.

src/users/entitlements/EntitlementForm.jsx
```diff
@@ -0,0 +1,108 @@
...
+      return;
+      break;
+  }
+}
+
+EntitlementForm.propTypes = {
```
awaisdar001, line 53: ideally, `entitlement` value should be reused.

src/users/entitlements/EntitlementForm.jsx
```diff
@@ -0,0 +1,108 @@
...
+    user: PropTypes.string.isRequired,
+  }),
+  user: PropTypes.string.isRequired,
+  changeHandler: PropTypes.func.isRequired,
+  closeHandler: PropTypes.func.isRequired,
+  forwardedRef: PropTypes.shape({ current: PropTypes.instanceOf(Element) }),
+};
+
+EntitlementForm.defaultProps = {
+  entitlement: {
```
awaisdar001: same goes here.

src/users/entitlements/EntitlementReissueForm.jsx
```diff
@@ -0,0 +1,142 @@
...
+  const [courseUuid, setCourseUuid] = useState(entitlement.courseUuid);
+  const [mode, setMode] = useState(entitlement.mode);
+  const [comments, setComments] = useState('');
+  const { add, clear } = useContext(UserMessagesContext);
+
+  const submit = useCallback(() => {
+    clear('entitlements');
+    patchEntitlement({
+      uuid: entitlement.uuid,
+      action: REISSUE,
```
awaisdar001: same comments here.

awaisdar001 (commented):
need some improvements here.

src/users/entitlements/EntitlementReissueForm.jsx
Ali-D-Akbar: What must be updated here? I've updated everything else except this.

src/users/entitlements/EntitlementForm.jsx
```diff
@@ -25,8 +26,7 @@ export default function EntitlementForm({
         forwardedRef={forwardedRef}
       />
     );
-  }
-  else if (formType === EXPIRE) {
+  } if (formType === EXPIRE) {
```
awaisdar001, line 28: Only one form is rendering at one time so what do you think about using `if/else`?

src/users/data/api.js
```diff
@@ -225,19 +225,18 @@ export async function patchEntitlement({
   action,
   unenrolledRun = null,
   comments = null,
+  expiredAt = null,
 }) {
   try {
     const { data } = await getAuthenticatedHttpClient().patch(
       `${getConfig().LMS_BASE_URL}/api/entitlements/v1/entitlements/${uuid}/`,
       {
```
awaisdar001: My understanding was when I commented previously, is to struct the following data from the calling function so API layer. 

```
  {
        expired_at: expiredAt,
        support_details: [{
          unenrolled_run: unenrolledRun,
          action,
          comments,
        }],
      },
```

src/users/entitlements/CreateEntitlementForm.jsx
```diff
@@ -28,38 +26,21 @@ export default function EntitlementForm({
...
+        result.errors.forEach(error => add(error));
+      } else {
+        changeHandler();
+      }
+    });
   });
-
-  const isReissue = formType === REISSUE;
-  const title = isReissue ? 'Re-issue Entitlement' : 'Create Entitlement';
+  const title = 'Create Entitlement';
```
awaisdar001: Any reason for having this variable in place? is being reused anywhere? If it's only used in one place then this variable is not useful.

src/users/entitlements/EntitlementForm.jsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
```
awaisdar001, line 1: the name of this file can be improved to `RenderForm.jsx`

src/users/entitlements/EntitlementPropTypes.jsx
```diff
@@ -0,0 +1,31 @@
+import PropTypes from 'prop-types';
```
awaisdar001: `src/users/entitlements/PropTypes.jsx`

src/users/entitlements/Entitlements.jsx
```diff
@@ -123,18 +124,32 @@ export default function Entitlements({
...
-          >
-            Reissue
-          </Button>
+          <div>
+            <Button
+              type="button"
+              disabled={!result.enrollmentCourseRun}
+              onClick={() => {
+                clearCourseSummary();
+                setUserEntitlement(result);
```
awaisdar001: `result` => `entitlement`

src/users/entitlements/ExpireEntitlementForm.jsx
```diff
@@ -0,0 +1,117 @@
...
+}) {
+  const [courseUuid, setCourseUuid] = useState(entitlement.courseUuid);
+  const [mode, setMode] = useState(entitlement.mode);
+  const [comments, setComments] = useState('');
+  const { add, clear } = useContext(UserMessagesContext);
+
+  const submit = useCallback(() => {
+    const now = new Date().toISOString();
+    clear('entitlements');
+    patchEntitlement({
```
awaisdar001, line 29:
```
patchEntitlement({
uuid: entitlement.uuid,
data:  {
        expired_at: now,
        support_details: [{
          unenrolled_run: entitlement.enrollmentCourseRun,
          action: EXPIRE,
          comments,
        }],
      })
....
```

src/users/entitlements/EntitlementForm.jsx
Ali-D-Akbar: Had to change it because of eslint error `no-else-return`
(It's weird how eslint reacts to 3 cases of if/else.)

src/users/data/api.js
Ali-D-Akbar: Yes, this is what I did.

src/users/entitlements/CreateEntitlementForm.jsx
Ali-D-Akbar: You're right. I'm removing the variable and use the text directly.

src/users/entitlements/ExpireEntitlementForm.jsx
Ali-D-Akbar: Implemented. 👍

src/users/data/api.js
```diff
@@ -221,24 +221,11 @@ export async function getCourseData(courseUUID) {
 }
 
 export async function patchEntitlement({
```
awaisdar001, line 223: this API call is alot cleaner now. 👍 
have you updated all `patchEntitlement` calls in the code?

src/users/entitlements/Entitlements.jsx
```diff
@@ -65,76 +66,90 @@ export default function Entitlements({
     if (data === null) {
       return [];
     }
-    return data.results.map(result => ({
+    return data.results.map(entitlement => ({
```
awaisdar001, line 69: more readable with this change 👍

awaisdar001 (approved):
the changes look good to me 👍

src/users/data/api.js
Ali-D-Akbar: Yes, I've changed the calls accordingly and tested it on the local.

src/users/entitlements/RenderForm.jsx
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { CREATE, REISSUE, EXPIRE } from './EntitlementActions';
+import CreateEntitlementForm from './CreateEntitlementForm';
+import ReissueEntitlementForm from './ReissueEntitlementForm';
+import ExpireEntitlementForm from './ExpireEntitlementForm';
+import { EntitlementPropTypes, EntitlementDefaultProps } from './PropTypes';
+
+export default function RenderForm({
```
DawoudSheraz: I think the appropriate name would be EntitlementForm. RenderForm doesn't signify what is being rendered.

src/users/entitlements/ReissueEntitlementForm.jsx
```diff
@@ -0,0 +1,116 @@
...
+    clear('entitlements');
+    patchEntitlement({
+      uuid: entitlement.uuid,
+      requestData: {
+        support_details: [{
+          unenrolled_run: entitlement.enrollmentCourseRun,
+          action: REISSUE,
+          comments,
+        }],
+      },
```
DawoudSheraz: I think we can write a utility function that will return the dict representation of the data that API needs. Right now, the representation is being duplicated whenever we need to call API. Although this will not go to the API function itself, having a utility function means that in the future, if the API data signature changes, we will need to update it in fewer places.

src/users/entitlements/Entitlements.jsx
```diff
@@ -65,76 +66,90 @@ export default function Entitlements({
...
+                setUserEntitlement(entitlement);
+                setFormType(REISSUE);
+              }}
+            >
+              Reissue
+            </Button>
+            <Button
+              type="button"
+              variant="outline-danger"
+              disabled={entitlement.expiredAt}
```
DawoudSheraz, line 143: Isn't expiredAt a datetime str?

DawoudSheraz (commented):
Added a few comments.

src/users/entitlements/RenderForm.jsx
Ali-D-Akbar: I originally had it named as `EntitlementForm` but was asked to change it by @awaisdar001

src/users/entitlements/Entitlements.jsx
Ali-D-Akbar: Yes, if `expiredAt`HAS a value other than `null` then the field will be disabled.

src/users/entitlements/ReissueEntitlementForm.jsx
DawoudSheraz: @Ali-D-Akbar

src/users/entitlements/RenderForm.jsx
DawoudSheraz: Hmm, I still think EntitlementForm is a better name. It signifies the form is for entitlement. RenderForm is a generic term because it could be to render anything.

DawoudSheraz (commented):
nit comments, otherwise it is good to approve.

src/users/entitlements/ReissueEntitlementForm.jsx
Ali-D-Akbar: Updated.

src/users/entitlements/ExpireEntitlementForm.jsx
```diff
@@ -0,0 +1,117 @@
...
+            className={classNames(
+              'btn-primary mr-3',
+              { disabled: !(courseUuid && mode && comments) },
+            )}
+            onClick={submit}
+          >
+            Submit
+          </Button>
+          <Button
+            className="btn-outline-secondary"
```
DawoudSheraz: use variant here. It might break with paragon 12. Not a blocker but do change it in a followup PR

src/users/entitlements/ReissueEntitlementForm.jsx
```diff
@@ -0,0 +1,115 @@
...
+            className={classNames(
+              'btn-primary mr-3',
+              { disabled: !(courseUuid && mode && comments) },
+            )}
+            onClick={submit}
+          >
+            Submit
+          </Button>
+          <Button
+            className="btn-outline-secondary"
```
DawoudSheraz: same here

src/users/entitlements/makeRequestData.js
```diff
@@ -0,0 +1,12 @@
+export default function makeRequestData({
```
DawoudSheraz: filename should be utils.js, not makeRequestData

DawoudSheraz (approved):

PR #88, opened 2021-01-27
fix long course_id breaking enrollment table

DawoudSheraz:
### Description
Fix the table overflowing due to long course id by adding word-break on course_id a tag

#### Before
Screenshot 2021-01-27 at 6 56 45 PM

#### After
Screenshot 2021-01-27 at 6 56 55 PM

### DIFF
--- a/src/users/index.scss
+++ b/src/users/index.scss
@@ -5,3 +5,7 @@ button.toggle-password {
 button.neg-margin-top{
   margin-top: -7px;
 }
+
+a.word_break {
+word-break:break-word;
+}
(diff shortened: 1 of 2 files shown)


src/users/index.scss
```diff
@@ -5,3 +5,7 @@ button.toggle-password {
 button.neg-margin-top{
   margin-top: -7px;
 }
+
+a.word_break {
+word-break:break-word;
```
awaisdar001: nit: alignment.
2 spaces here.

awaisdar001 (approved):

PR #89, opened 2021-01-28
convert all buttons to use variant attr

DawoudSheraz:
### Description
Fix the buttons to use `variant` attribute instead of using btn- classes.

Screenshot 2021-01-28 at 7 03 26 PM
Screenshot 2021-01-28 at 7 03 07 PM
Screenshot 2021-01-28 at 6 57 13 PM


awaisdar001 (approved):

saadyousafarbi (approved):
LGTM!

PR #95, opened 2021-03-12
api calls tests

DawoudSheraz:
### Description
Adds the unit tests for API calls methods in support tools. A new dependency, [axios-mock-adapter](https://www.npmjs.com/package/axios-mock-adapter), has been added to mock the api calls and responses. To mock the auth user, [MockAuthService](https://github.com/edx/frontend-platform/blob/master/src/auth/MockAuthService.js) from edx/frontend-platform has been configured. This mock service allows to add an authenticated user and then mock the API calls made via the authenticated user.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5798,6 +5798,24 @@
         "lodash": "^4.17.11"
       }
     },
+    "axios-mock-adapter": {
+      "version": "1.19.0",
+      "resolved": "https://registry.npmjs.org/axios-mock-adapter/-/axios-mock-adapter-1.19.0.tgz",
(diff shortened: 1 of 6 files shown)


awaisdar001 (approved):
the changes looks good to me. Just a question -- how much coverage are you targeting for the api.js file?

DawoudSheraz:
> the changes looks good to me. Just a question -- how much coverage are you targeting for the api.js file?

Almost 100%. The one missed case will never be hit in the code flow. That redundant code(along with some others) will be removed in another ticket.

awaisdar001:
sounds good. 👍

azanbinzahid (approved):

PR #98, opened 2021-03-25
entitlement tests

DawoudSheraz:
### Description
Added unit tests for entitlements directory components. Besides the addition of unit tests, disabled check in the entitlement forms has been added. The previous check was only adding the disabled class but not affecting the behavior.

Coverage: 64%

### DIFF
--- a/src/users/entitlements/CreateEntitlementForm.test.jsx
+++ b/src/users/entitlements/CreateEntitlementForm.test.jsx
@@ -1,0 +1,80 @@
+import { mount } from 'enzyme';
+import { waitFor } from '@testing-library/react';
+import React from 'react';
+
+import { waitForComponentToPaint } from '../../setupTest';
+import CreateEntitlementForm from './CreateEntitlementForm';
+import entitlementFormData from '../data/test/entitlementForm';
+import UserMessagesProvider from '../../user-messages/UserMessagesProvider';
+import * as api from '../data/api';
+
+const CreateEntitlementFormWrapper = (props) => (
+  <UserMessagesProvider>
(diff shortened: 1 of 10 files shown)


src/users/entitlements/CreateEntitlementForm.test.jsx
```diff
@@ -0,0 +1,80 @@
...
+    it('Submit button disabled by default', () => {
+      expect(wrapper.find('button.btn-primary').prop('disabled')).toBeTruthy();
+    });
+
+    it('Successful form submission', async () => {
+      const apiMock = jest.spyOn(api, 'postEntitlement').mockImplementationOnce(() => Promise.resolve({}));
+      expect(apiMock).toHaveBeenCalledTimes(0);
+
+      wrapper.find('textarea#comments').simulate('change', { target: { value: 'creating new entitlement' } });
+      const submitButton = wrapper.find('button.btn-primary');
```
awaisdar001, line 47: initialization of `submitButton` should go in the top block.

awaisdar001 (approved):
Looks good to me 👍

Ali-D-Akbar (approved):
LGTM.

src/users/entitlements/CreateEntitlementForm.test.jsx
DawoudSheraz: actually, this causes an error. Once the comments are added, only then the submit button state updates. I can move the initialization above but this line will still remain here.

awaisdar001: got it. makes sense.

PR #102, opened 2021-04-01
Add Licenses Subscription Data for Users

azanbinzahid:
## Description
This PR adds `Licenses` table component  which fetch user license subscription data from `edx/license-manager` and display in User page along with other user related data. Relevant CORS, stage and prod config has already been added edx/license-manager#207, edx/edx-internal#4844

## Linked issue ticket
[PROD-2257](https://openedx.atlassian.net/browse/PROD-2257)

## Local Dependency
`edx/license-manager` should be running in devstack along with a subscription added against LMS user

## Screenshot
image

## Testing
Full functionality, possible errors and connectivity must be tested on stage before deploying to prod

## Commit Squash
Commits will be squashed into one with `feat: add user license data from license-manager` message before merge


src/users/UserPage.jsx
```diff
@@ -149,6 +154,11 @@ export default function UserPage({ location }) {
             ssoRecords={data.ssoRecords}
             changeHandler={handleUserSummaryChange}
           />
+          <Licenses
+            data={data.licenses.results}
+            status={data.licenses.status}
+            expanded={showLicenses}
+          />
```
DawoudSheraz, line 161: Add a related condition in UserPage.test.jsx successful page load test(verify that this element is rendered)

src/users/data/api.js
```diff
@@ -156,13 +156,53 @@ export async function getUserPasswordStatus(userIdentifier) {
...
+    status: '',
+    results: [],
+  };
+  try {
+    const { data } = await getAuthenticatedHttpClient().post(
+      `${getConfig().LICENSE_MANAGER_URL}/api/v1/staff_lookup_licenses/`,
+      { user_email: userEmail },
+    );
+    defaultResponse.results = data;
+    defaultResponse.status = 'Successful';
```
DawoudSheraz: For success, don't show any status at all. This info will be helpful for the failure cases only.

src/users/Licenses.jsx
```diff
@@ -0,0 +1,138 @@
...
+  ];
+
+  const tableDataSortable = [...tableData];
+  return (
+    <section className="mb-3">
+      <Collapsible
+        title={(
+          <>
+            {`Licenses (${tableData.length})`}
+            <Badge variant="light">{`Fetch Status: ${responseStatus}`}</Badge>
```
DawoudSheraz: Show the status if there is an error fetching the results. No need to show success status(redundant given that data request call has return successfully and tableData.length signifies how many data elements have been returned).

DawoudSheraz (changes requested):
added a few minor comments to address

src/users/data/api.js
azanbinzahid: Code updated

src/users/Licenses.jsx
azanbinzahid: updated

src/users/UserPage.jsx
azanbinzahid: Having difficulty in understanding this, can you elaborate?

src/users/Licenses.jsx
```diff
@@ -0,0 +1,145 @@
...
+      label: 'Activation Link', key: 'activationLink', columnSortable: true, onSort: () => setSort('activationLink'), width: 'col-3',
+    },
+  ];
+
+  const tableDataSortable = [...tableData];
+  let statusMsg;
+  if (responseStatus !== '') {
+    statusMsg = <Badge variant="light">{`Fetch Status: ${responseStatus}`}</Badge>;
+  } else {
+    statusMsg = <></>;
```
DawoudSheraz: Use React Fragment to create empty element

DawoudSheraz (approved):
a nit to address before merging.

src/users/Licenses.jsx
```diff
@@ -0,0 +1,145 @@
...
+import { Collapsible, Badge } from '@edx/paragon';
+import sort from './sort';
+import Table from '../Table';
+import formatDate from '../dates/formatDate';
+
+export default function Licenses({
+  data, status, expanded,
+}) {
+  const [sortColumn, setSortColumn] = useState('status');
+  const [sortDirection, setSortDirection] = useState('desc');
```
awaisdar001, line 16: This can be a good candidate for a boolean value. 
```
const [sortInAscendingOrder, setSortInAscendingOrder] = useState(False);
```

src/users/Licenses.jsx
```diff
@@ -0,0 +1,145 @@
...
+      activationLink: {
+        displayValue: <a href={result.activationLink} rel="noopener noreferrer" target="_blank" className="word_break">{result.activationLink}</a>,
+        value: result.activationLink,
+      },
+    }));
+  }, [data]);
+
+  const setSort = useCallback((column) => {
+    if (sortColumn === column) {
+      setSortDirection(sortDirection === 'desc' ? 'asc' : 'desc');
```
awaisdar001, line 58: this could become 
```
if (sortColumn === column) {
    setSortInAscendingOrder(!sortInAscendingOrder) 
} else {
    setSortDirection(False);
}
```

src/users/Licenses.jsx
awaisdar001: I think this is the newer way to create an empty component that is identical to React.Fragment. The only difference is that older versions of the JSX Babel plugin didn’t understand <>, so if you need to support those older versions you need to use <Fragment>, otherwise it makes no difference.

awaisdar001 (approved):
👍 Nice work -- Just wanted to flag that my comments just for info and are not blockers.

src/users/Licenses.test.jsx
```diff
@@ -0,0 +1,42 @@
...
+    wrapper = mount(<Licenses {...licenseData} />);
+    const collapsible = wrapper.find('CollapsibleAdvanced').find('.collapsible-trigger').hostNodes();
+    expect(collapsible.text()).toEqual('Licenses (0)Fetch Status: No record found');
+  });
+
+  it('Sorting Columns Button Enabled by default', () => {
+    const dataTable = wrapper.find('table.table');
+    const tableHeaders = dataTable.find('thead tr th');
+
+    tableHeaders.forEach(header => {
```
Ali-D-Akbar, line 37: There is some coverage that is being dropped in Licenses.jsx and It can be covered by checking the labels for each row.

src/users/data/test/licenses.js
```diff
@@ -0,0 +1,28 @@
+const licensesData = {
+  data: [
+    {
+      status: 'unassigned',
+      assignedDate: null,
+      revokedDate: null,
+      activationDate: '2021-04-01',
```
Ali-D-Akbar: I think you should use `Date().toLocaleString()` instead of manually adding a date.

src/users/Licenses.jsx
azanbinzahid: I think using `null` would also work here.

src/users/Licenses.jsx
azanbinzahid: Yes, this would simplify the logic here but then we would still need to convert this boolean to `asc` and `desc` later when passing props to `Table` component.

src/users/data/test/licenses.js
azanbinzahid: The API actually is returning a date with dashes in between, and `Date().toLocaleString()` is returning in some other format.

Ali-D-Akbar: Ohh, I guess you should preferably try some other formatting functions.

src/users/Licenses.test.jsx
azanbinzahid: I have tried to get example from other components but unable to find, may be we can address this in separate PR where we will add tests for table in all components.

Ali-D-Akbar: I recently added something similar in the tests here: https://github.com/edx/frontend-app-support-tools/blob/0760f3d9cb2e8a54b74d06f00205eaca3e664330/src/users/entitlements/Entitlements.test.jsx#L66

It checks the headings and covers these lines.

src/users/data/test/licenses.js
azanbinzahid: `new Date().toISOString().replace(/T.*/,'')` this worked

src/users/Licenses.test.jsx
azanbinzahid: Just added a test for header length

Ali-D-Akbar (approved):

PR #106, opened 2021-04-06
Add Reset Password Feature

Ali-D-Akbar:
A new `Reset Password` button has been added in Account section:
image

Upon clicking the button, a confirmation modal will be shown. Clicking the `Confirm` button will then send an email to the respective user:
image

### DIFF
--- a/src/users/UserSummary.jsx
+++ b/src/users/UserSummary.jsx
@@ -2,7 +2,7 @@ import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import { Modal, Button, Input } from '@edx/paragon';
 
-import { postTogglePasswordStatus } from './data/api';
+import { postTogglePasswordStatus, postResetPassword } from './data/api';
 import Table from '../Table';
 import formatDate from '../dates/formatDate';
 
@@ -16,6 +16,7 @@ export default function UserSummary({
   const [idvModalIsOpen, setIdvModalIsOpen] = useState(false);
   const [disableUserModalIsOpen, setDisableUserModalIsOpen] = useState(false);
   const [disableHistoryModalIsOpen, setDisableHistoryModalIsOpen] = useState(false);
+  const [resetPasswordModalIsOpen, setResetPasswordModalIsOpen] = useState(false);
   const [comment, setComment] = useState('');
   const [userPasswordHistoryData, setUserPasswordHistoryData] = useState([]);
   const [extraSsoDataTitle, setSsoExtraDataTitle] = useState('');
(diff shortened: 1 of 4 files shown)


src/users/UserSummary.jsx
```diff
@@ -254,13 +260,19 @@ export default function UserSummary({
                 >
                   {userData.passwordStatus.status === PASSWORD_STATUS.USABLE ? 'Disable User' : 'Enable User'}
                 </Button>
+                <Button
+                  className="reset-password"
```
DawoudSheraz: unless there is a related css class, use id instead of class.

src/users/UserSummary.jsx
```diff
@@ -348,6 +360,29 @@ export default function UserSummary({
             </div>
           )}
         />
+        <Modal
+          open={resetPasswordModalIsOpen}
+          id="user-account-reset-password"
+          buttons={[
+            <Button
+              variant="danger"
+              onClick={resetPassword}
```
DawoudSheraz, line 369: Why resetPassword call is this way?

Ali-D-Akbar: Should I call `postResetPassword(userData.email)` right away?

DawoudSheraz: no, not that. I am asking about the format. Why () => resetPassword is not used? Is it an ES6 or linting thing?

Ali-D-Akbar: `resetPassword` itself is an arrow function. That's why we don't need to write another arrow function which will call `resetPassword`.

We usually use arrow functions in such events if they're one liner or short. E.G:

`onclick={() => setResetPasswordModalIsOpen(true)}`

here we don't even need to write `return` keyword if it is one liner.

DawoudSheraz (approved):

src/users/UserSummary.jsx
```diff
@@ -35,6 +36,11 @@ export default function UserSummary({
     changeHandler();
   };
 
+  const resetPassword = () => {
+    postResetPassword(userData.email);
```
azanbinzahid, line 40: Will this `userData` always have email in it or we need add some checks?

src/users/data/api.js
```diff
@@ -332,3 +332,13 @@ export async function postTogglePasswordStatus(user, comment) {
...
 }
+
+export async function postResetPassword(email) {
+  const { data } = await getAuthenticatedHttpClient().post(
+    `${getConfig().LMS_BASE_URL}/account/password`,
+    {
+      email,
+    },
+  );
+  return data;
```
azanbinzahid, line 343: Shouldn't we need to log return status or any error (if returned)?

src/users/UserSummary.jsx
Ali-D-Akbar: Every user will have an email associated with it since edX always stores users with the emails.
userData will always have all the info about a user.

src/users/data/api.js
Ali-D-Akbar: I don't think so, in this case, the API only returns status 200 with an empty response.

azanbinzahid (approved):

PR #110, opened 2021-04-21
refactor: Code Cleanup

Ali-D-Akbar:
[PROD-2222](https://openedx.atlassian.net/browse/PROD-2222)
The refactoring contains the following changes:
1. Made a separate centralized URLs file containing all the api endpoint links
2. Enforced eslint rules wherever applicable and disabling the rest
3. Shifting util/helper functions in a separate file `utils.js`
4. Used `variant` attribute instead of `className` for all paragon components
5. Removed redundant and unnecessary code
6. Improve tests based on usage
7. Modularized and moved components in separate folders such as entitlement, enrollment, course summary
8. Renamed a lot of files and folders


src/users/CourseSummary.jsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import PageLoading from '../components/common/PageLoading';
-import formatDate from '../dates/formatDate';
+import { formatDate } from '../utils/index';
```
awaisdar001: this should work as well. 
`import { formatDate } from '../utils';`

src/users/data/urls.js
```diff
@@ -0,0 +1,69 @@
+import { getConfig } from '@edx/frontend-platform';
+import { isEmail, isValidUsername } from '../../utils/index';
+
+export const GET_ENTITLEMENTS_URL = () => `${
```
awaisdar001: @Ali-D-Akbar any particular reason why these function names are in captial? 

I was expecting more like the following. 
```
getEntitlementUrl = () => `${ ...

getEnrollmentUrl = (username) => ...

```

Ali-D-Akbar: I mixed the two concepts (capitalizing the constants and arrow functions)
I'll change this to camel lowercase. 👍

src/utils/index.test.js
```diff
@@ -1,32 +1,46 @@
...
-    'www.email.com',
-  ];
-  const validUsername = ['staff', 'admin123'];
-  const invalidUsername = ['', ' ', undefined, 'invalid username', '%invalid'];
-  test.each(validEmails)('isEmail return true for %s', (email) => {
-    expect(isEmail(email)).toBe(true);
-  });
-  test.each(invalidEmails)('isEmail return false for %s', (email) => {
-    expect(isEmail(email)).toBe(false);
+  describe('isEmail isValidUsername', () => {
```
DawoudSheraz: nit: test user identifier.

src/users/course-summary/CourseSummary.test.jsx
```diff
@@ -2,7 +2,7 @@ import { mount } from 'enzyme';
 import React from 'react';
 
 import CourseSummary from './CourseSummary';
-import CourseSummaryData from './data/test/courseSummary';
+import CourseSummaryData from '../data/test/courseSummary';
```
DawoudSheraz: use courseSummary as dir name instead of course-summary

azanbinzahid:
Can we use absolute paths in imports rather then relative?

src/users/data/api.js
```diff
@@ -1,12 +1,14 @@
-import { getConfig, ensureConfig } from '@edx/frontend-platform';
+import { ensureConfig } from '@edx/frontend-platform';
 import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
 import * as messages from '../../user-messages/messages';
-import { isEmail, isValidUsername } from '../../utils/index';
+import {
+  getEntitlementsUrl, getEnrollmentsUrl, getSsoRecordsUrl, getUserUrl, getUserVerificationDetailUrl,
```
awaisdar001: how about this syntax - looks cleaner. 
```
import * as AppUrls from './urls'
...
...
AppUrls.getEnrollmentsUrl(username)

```

src/users/data/urls.js
```diff
@@ -0,0 +1,69 @@
...
+
+export const getEnrollmentsUrl = (username) => `${
+  getConfig().LMS_BASE_URL
+}/support/enrollment/${username}`;
+
+export const getSsoRecordsUrl = (username) => `${
+  getConfig().LMS_BASE_URL
+}/support/sso_records/${username}`;
+
+export const getUserUrl = (userIdentifier) => {
```
awaisdar001: This name should be updated to indicate what kind of user url it is. 
e.g. `user account URL` or `user preferences URL`

src/users/data/api.js
```diff
@@ -41,18 +43,7 @@ export async function getSsoRecords(username) {
...
-
-  // todo: we have already validated the input to fall into one of these cases.
-  // The following `if` is not required.
-  if (!(identifierIsEmail || identifierIsUsername)) {
-    throw new Error('Invalid Argument!');
-  }
-  url = identifierIsEmail
-    ? (url += `?email=${userIdentifier}`)
-    : (url += `/${userIdentifier}`);
+  const [url, identifierIsEmail] = getUserUrl(userIdentifier);
```
awaisdar001: Other than the name change, this doesn't need to send whether or not `userIdentifier` was email or not just because we need it several lines below. 

Let this function only return the `kindOfUserUrl` (rename `url` const).

src/users/data/api.js
```diff
@@ -41,18 +43,7 @@ export async function getSsoRecords(username) {
...
 
   try {
     const { data } = await getAuthenticatedHttpClient().get(url);
```
awaisdar001, line 46: Let's move the logic of selecting the error message at L#57 to a util method. The util method should know the logic to select an error message based on `userIdentifer`.

awaisdar001: Secondly, the code for `error.userError` is being repeated - let's think about how we can improve it. 
```
error.userError = {
        code: null,
        dismissible: true,
        text: notFoundErrorText,
        type: 'error',
        topic: 'general',
      };
```

src/users/data/api.js
```diff
@@ -98,9 +89,7 @@ export async function getUserVerificationDetail(username) {
   };
   try {
     const { data } = await getAuthenticatedHttpClient().get(
-      `${
-        getConfig().LMS_BASE_URL
-      }/api/user/v1/accounts/${username}/verifications/`,
+      getUserVerificationDetailUrl(username),
```
awaisdar001: Much cleaner function name. +1

src/users/data/api.js
```diff
@@ -167,7 +151,7 @@ export async function getLicense(userEmail) {
   };
   try {
     const { data } = await getAuthenticatedHttpClient().post(
-      `${getConfig().LICENSE_MANAGER_URL}/api/v1/staff_lookup_licenses/`,
+      getLicenseUrl(),
```
awaisdar001: `getLicenseManagerUrl`

src/users/data/api.js
```diff
@@ -266,7 +250,7 @@ export async function patchEntitlement({
 }) {
   try {
     const { data } = await getAuthenticatedHttpClient().patch(
-      `${getConfig().LMS_BASE_URL}/api/entitlements/v1/entitlements/${uuid}/`, requestData,
+      patchEntitlementUrl(uuid), requestData,
```
awaisdar001: Rather than having two different names `patchEntitlementUrl` and `postEntitlementUrl`, please think about how we can achive this from one method and call it `getEntitlementUrl`. 

I can only see one difference in the url which is uuid. I would have gone for one method with if condition. 
e.g.
```
export const getEntitlementUrl = (uuid) => {
    const baseUrl =`${getConfig().LMS_BASE_URL}/api/entitlements/v1/entitlements/`;
    return uuid ? `${baseUrl}{uuid}/` : baseUrl
}
```

awaisdar001: btw `get` doesn't mean the type of request in this context.

src/users/data/api.js
```diff
@@ -332,7 +316,7 @@ export async function postEnrollmentChange({
 }) {
   try {
     const { data } = await getAuthenticatedHttpClient().post(
-      `${getConfig().LMS_BASE_URL}/support/enrollment/${user}`,
+      postEnrollmentChangeUrl(user),
```
awaisdar001: => `getEnrollmentChangeUrl`
do not confuse function naming with the request types.

src/users/data/api.js
```diff
@@ -366,7 +350,7 @@ export async function postEnrollmentChange({
 
 export async function postTogglePasswordStatus(user, comment) {
   const { data } = await getAuthenticatedHttpClient().post(
-    `${getConfig().LMS_BASE_URL}/support/manage_user/${user}`,
+    postTogglePasswordStatusUrl(user),
```
awaisdar001: `getTogglePasswordStatusUrl`

awaisdar001 (commented):
I added few questions and few recommendations, totally optional and are not blockers :)

src/users/data/api.js
Ali-D-Akbar: Except for this, I've made changes according to your suggestions.
This code is just used in 2 places. I think it can be left unchanged.

Ali-D-Akbar:
Suggested changes have been made.

DawoudSheraz (approved):

azanbinzahid:
> Can we use absolute paths in imports rather then relative?

@Ali-D-Akbar just a small comment.

Ali-D-Akbar:
> Can we use absolute paths in imports rather then relative?

@azanbinzahid  It's easier to use absolute imports from an app created from create-react-app. For this, we don't have any plans to implement it yet since it doesn't use create-react-app

azanbinzahid (approved):

awaisdar001 (approved):
:shipit:

PR #113, opened 2021-05-04
refactor: add test coverage to various files

Ali-D-Akbar:
[PROD-2354](https://openedx.atlassian.net/browse/PROD-2354)
Adding and changing some tests and files to increase test coverage for Support Tools.

### DIFF
--- a/src/users/UserPage.test.jsx
+++ b/src/users/UserPage.test.jsx
@@ -54,6 +54,17 @@ describe('User Page', () => {
         throw error;
       });
     };
+    it('when user identifier is empty', () => {
+      const emptyUsername = '';
+      location.search = `?username=${emptyUsername}`;
+      const wrapper = mount(<UserPageWrapper location={location} />);
+
+      const searchInput = wrapper.find('input[name="userIdentifier"]');
+      const alert = wrapper.find('.alert');
+
+      expect(searchInput).toHaveLength(1);
+      expect(alert).toHaveLength(0);
+    });
     it('when user identifier is invalid', () => {
       const invalidUsername = 'invalid username';
       location.search = `?username=${invalidUsername}`;
@@ -81,6 +92,7 @@ describe('User Page', () => {
 
       const alert = wrapper.find('.alert');
       expect(alert).toHaveLength(1);
+      expect(document.title).toEqual('Support Tools | edX');
       expect(alert.text()).toEqual(expectedAlert);
     });
     it('when user email is not found', async () => {
--- a/src/users/enrollments/Enrollments.test.jsx
+++ b/src/users/enrollments/Enrollments.test.jsx
@@ -51,4 +51,17 @@ describe('Course Enrollments Listing', () => {
       expect(sortButton.disabled).toBeFalsy();
     });
   });
+
+  it('Sorting Columns Button work correctly', () => {
+    const dataTable = wrapper.find('table.table');
+    const tableHeaders = dataTable.find('thead tr th');
+
+    tableHeaders.forEach(header => {
+      const sortButton = header.find('button.btn-header');
+      sortButton.simulate('click');
+      expect(wrapper.find('svg.fa-sort-down')).toHaveLength(1);
+      sortButton.simulate('click');
+      expect(wrapper.find('svg.fa-sort-up')).toHaveLength(1);
+    });
+  });
 });
--- a/src/users/entitlements/CreateEntitlementForm.test.jsx
+++ b/src/users/entitlements/CreateEntitlementForm.test.jsx
@@ -43,6 +43,8 @@ describe('Create Entitlement Form', () => {
       const apiMock = jest.spyOn(api, 'postEntitlement').mockImplementationOnce(() => Promise.resolve({}));
       expect(apiMock).toHaveBeenCalledTimes(0);
 
+      wrapper.find('input#courseUuid').simulate('change', { target: { value: 'b4f19c72-784d-4110-a3ba-318666a7db1a' } });
+      wrapper.find('select#mode').simulate('change', { target: { value: 'professional' } });
       wrapper.find('textarea#comments').simulate('change', { target: { value: 'creating new entitlement' } });
       const submitButton = wrapper.find('button.btn-primary');
       expect(submitButton.prop('disabled')).toBeFalsy();
--- a/src/users/entitlements/Entitlements.test.jsx
+++ b/src/users/entitlements/Entitlements.test.jsx
@@ -48,6 +48,28 @@ describe('Entitlements Listing', () => {
     expect(collapsible.text()).toEqual('Entitlements (0)');
   });
 
+  it('Sorting Columns Button Enabled by default', () => {
+    const dataTable = wrapper.find('table.table');
+    const tableHeaders = dataTable.find('thead tr th');
+
+    tableHeaders.forEach(header => {
+      const sortButton = header.find('button.btn-header');
+      expect(sortButton.disabled).toBeFalsy();
+    });
+  });
+
+  it('Sorting Columns Button work correctly', () => {
+    const dataTable = wrapper.find('table.table').hostNodes();
+    const sortButtons = dataTable.find('thead tr th button.btn-header');
+
+    sortButtons.forEach(sortButton => {
+      sortButton.simulate('click');
+      expect(wrapper.find('svg.fa-sort-down')).toHaveLength(1);
+      sortButton.simulate('click');
+      expect(wrapper.find('svg.fa-sort-up')).toHaveLength(1);
+    });
+  });
+
   it('Support Details data', () => {
     wrapper = mount(<EntitlementsPageWrapper {...entitlementsData} />);
     const tableRowsLengths = [2, 0];
--- a/src/users/licenses/Licenses.jsx
+++ b/src/users/licenses/Licenses.jsx
@@ -114,7 +114,6 @@ export default function Licenses({
           )}
           columns={columns}
           tableSortable
-          defaultSortedColumn="status"
           defaultSortDirection="desc"
         />
       </Collapsible>
--- a/src/utils/index.test.js
+++ b/src/utils/index.test.js
@@ -1,4 +1,6 @@
-import { isEmail, isValidUsername, formatDate } from './index';
+import {
+  isEmail, isValidUsername, formatDate, sort,
+} from './index';
 
 describe('Test Utils', () => {
   describe('test user identifier', () => {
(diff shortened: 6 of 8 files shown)


src/users/enrollments/Enrollments.test.jsx
```diff
@@ -51,4 +51,15 @@ describe('Course Enrollments Listing', () => {
...
   });
+
+  it('Sorting Columns Button work correctly', () => {
+    const dataTable = wrapper.find('table.table');
+    const tableHeaders = dataTable.find('thead tr th');
+
+    tableHeaders.forEach(header => {
+      const sortButton = header.find('button.btn-header');
+      sortButton.simulate('click');
+      sortButton.simulate('click');
```
DawoudSheraz: What is being tested here? What happens when the button is clicked?

src/users/entitlements/EntitlementForm.jsx
```diff
@@ -14,8 +14,9 @@ export default function EntitlementForm({
   user,
   forwardedRef,
 }) {
+  let form = null;
```
DawoudSheraz: why is this change done?

src/users/entitlements/EntitlementForm.test.jsx
```diff
@@ -35,4 +35,12 @@ describe('Entitlement forms', () => {
...
     expect(wrapper.find('ReissueEntitlementForm').length).toEqual(0);
   });
+
+  it('Does not render any form', () => {
+    wrapper = mount(<EntitlementFormWrapper {...entitlementFormData} formType="" />);
+    expect(wrapper).toEqual({});
+    expect(wrapper.find('ExpireEntitlementForm').length).toEqual(0);
+    expect(wrapper.find('CreateEntitlementForm').length).toEqual(0);
+    expect(wrapper.find('ReissueEntitlementForm').length).toEqual(0);
+  });
```
DawoudSheraz: Why is this test needed?

src/users/entitlements/Entitlements.test.jsx
```diff
@@ -48,6 +48,27 @@ describe('Entitlements Listing', () => {
     expect(collapsible.text()).toEqual('Entitlements (0)');
   });
 
+  it('Sorting Columns Button Enabled by default', () => {
+    const dataTable = wrapper.find('table.w-100');
```
DawoudSheraz: use table.table

src/users/entitlements/Entitlements.test.jsx
```diff
@@ -48,6 +48,27 @@ describe('Entitlements Listing', () => {
...
+
+  it('Sorting Columns Button work correctly', () => {
+    const dataTable = wrapper.find('table.w-100');
+    const tableHeaders = dataTable.find('thead tr th');
+
+    tableHeaders.forEach(header => {
+      const sortButton = header.find('button.btn-header');
+      sortButton.simulate('click');
+      sortButton.simulate('click');
+    });
```
DawoudSheraz: Again, what is the effect of clicking sort button? Also, why 2 clicks?

src/utils/index.test.js
```diff
@@ -43,4 +45,29 @@ describe('Test Utils', () => {
...
+    const dict1 = {
+      name: 'string',
+      id: { value: 1 },
+      list: ['array value 1', 'array value 2'],
+    };
+    const dict2 = {
+      name: 'string2',
+      id: { value: 2 },
+      list: ['array value 3', 'array value 4'],
+    };
```
DawoudSheraz: use meaningful names as compared to dict1, dict2

DawoudSheraz (changes requested):
added some questions and comments to address

src/users/enrollments/Enrollments.test.jsx
Ali-D-Akbar: In order to check all possible cases for this code in [Enrollment.js](https://github.com/edx/frontend-app-support-tools/blob/master/src/users/enrollments/Enrollments.jsx#L84):

```
const setSort = useCallback((column) => {
    if (sortColumn === column) {
      setSortDirection(sortDirection === 'desc' ? 'asc' : 'desc');
    } else {
      setSortDirection('desc');
    }
    setSortColumn(column);
  });
```

  
both `simulate` calls must be done in order to call `setSortDirection`

src/users/entitlements/EntitlementForm.jsx
Ali-D-Akbar: It's a better practice to write if-else if instead of 3 if conditions. It also helps cover a couple more lines.

src/users/entitlements/EntitlementForm.test.jsx
Ali-D-Akbar: in order to increase test coverage.

src/users/entitlements/Entitlements.test.jsx
Ali-D-Akbar: it was causing problems as it may get more than 1 table component. Although I'll try this again.

src/users/enrollments/Enrollments.test.jsx
DawoudSheraz: Ok, separate out the tests(with asc and desc)

src/users/entitlements/EntitlementForm.test.jsx
DawoudSheraz: Hmm, if it doesn't give an actual advantage other than the coverage, don't include it.

src/users/entitlements/Entitlements.test.jsx
DawoudSheraz: use [hostNodes](https://enzymejs.github.io/enzyme/docs/api/ShallowWrapper/hostNodes.html)() if there is only 1 table but the result is both Table and table.

src/users/enrollments/Enrollments.test.jsx
Ali-D-Akbar: I think it'd be redundant to write another test as I'll be writing two variation of the buttons clicks (with 1 and the other with 2 button clicks)

src/users/entitlements/EntitlementForm.test.jsx
Ali-D-Akbar: Removed.

src/utils/index.test.js
Ali-D-Akbar: changed names.

src/users/entitlements/Entitlements.test.jsx
Ali-D-Akbar: Used another approach here as I was only receiving 2 table (not `Table`) elements.

src/users/enrollments/Enrollments.test.jsx
DawoudSheraz: No, it would not be redundant. Plus, add some checks to validate sorting has taken place. Just clicking does not indicate the success of sort.

src/users/entitlements/EntitlementForm.jsx
DawoudSheraz: Hmm, not a very compelling argument just to increase coverage.

src/users/entitlements/Entitlements.test.jsx
```diff
@@ -48,6 +48,29 @@ describe('Entitlements Listing', () => {
...
+  it('Sorting Columns Button work correctly', () => {
+    const dataTable = wrapper.find('table.table').hostNodes();
+    const tableHeaders = dataTable.find('thead tr th');
+
+    tableHeaders.forEach(header => {
+      const sortButton = header.find('button.btn-header');
+      if (sortButton.length === 1) {
+        sortButton.simulate('click');
+        sortButton.simulate('click');
+      }
```
DawoudSheraz: you don't add if conditions inside tests unnecessarily. Fetch the exact button via index and test appropriately. As for sort, again, test that sorting has actually worked(html structure changed etc.)

DawoudSheraz (changes requested):
some comments to address

src/users/entitlements/EntitlementForm.jsx
Ali-D-Akbar: Removing it.

src/users/enrollments/Enrollments.test.jsx
Ali-D-Akbar: I covered both cases in one test and added assertions in it.

src/users/entitlements/Entitlements.test.jsx
Ali-D-Akbar: removed if condition. My bad.

src/users/enrollments/Enrollments.test.jsx
```diff
@@ -51,4 +51,17 @@ describe('Course Enrollments Listing', () => {
...
+  it('Sorting Columns Button work correctly', () => {
+    const dataTable = wrapper.find('table.table');
+    const tableHeaders = dataTable.find('thead tr th');
+
+    tableHeaders.forEach(header => {
+      const sortButton = header.find('button.btn-header');
+      sortButton.simulate('click');
+      expect(wrapper.find('svg.fa-sort-down')).toHaveLength(1);
+      sortButton.simulate('click');
+      expect(wrapper.find('svg.fa-sort-up')).toHaveLength(1);
```
DawoudSheraz, line 64: nice

src/users/licenses/Licenses.jsx
```diff
@@ -114,7 +114,6 @@ export default function Licenses({
           )}
           columns={columns}
           tableSortable
-          defaultSortedColumn="status"
```
DawoudSheraz, line 117: why remove this?

DawoudSheraz (approved):

src/users/licenses/Licenses.jsx
Ali-D-Akbar: I thought of removing sort by default just like the other sortColumns to keep the consistency.

src/users/UserPage.test.jsx
```diff
@@ -81,6 +92,7 @@ describe('User Page', () => {
 
       const alert = wrapper.find('.alert');
       expect(alert).toHaveLength(1);
+      expect(document.title).toEqual('Support Tools | edX');
```
azanbinzahid, line 95: For future reference, we can define some site-wide constants e.g. Site title and we can just reference those anywhere needed.

src/users/entitlements/CreateEntitlementForm.test.jsx
```diff
@@ -43,6 +43,8 @@ describe('Create Entitlement Form', () => {
       const apiMock = jest.spyOn(api, 'postEntitlement').mockImplementationOnce(() => Promise.resolve({}));
       expect(apiMock).toHaveBeenCalledTimes(0);
 
+      wrapper.find('input#courseUuid').simulate('change', { target: { value: 'b4f19c72-784d-4110-a3ba-318666a7db1a' } });
```
azanbinzahid, line 46: where is this value coming from? can we use this value directly from some object instead of manually writing here?

azanbinzahid (approved):
added some questions but they are not blockers.

src/users/entitlements/CreateEntitlementForm.test.jsx
Ali-D-Akbar: It is just a dummy course uuid. We don't have any of these stored anywhere ATM.

src/users/UserPage.test.jsx
Ali-D-Akbar: Right. We can discuss this and get this done in the next installment of increasing coverage.

PR #119, opened 2021-05-17
feat: add activation link in user summary

azanbinzahid:
### Description
Adds activation link in user summary component. The field was added in user_api in https://github.com/edx/edx-platform/pull/27639

User with inactive status:
image

User with active status:
image

### DIFF
--- a/src/users/UserSummary.jsx
+++ b/src/users/UserSummary.jsx
@@ -1,10 +1,10 @@
 import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import { Modal, Button, Input } from '@edx/paragon';
-
 import { postTogglePasswordStatus, postResetPassword } from './data/api';
 import Table from '../Table';
 import { formatDate } from '../utils';
+import { getAccountActivationUrl } from './data/urls';
 
 export default function UserSummary({
   userData,
@@ -241,11 +241,28 @@ export default function UserSummary({
     },
   }));
 
+  if (!userData.isActive) {
+    let dataValue;
+    if (userData.activationKey !== null) {
+      dataValue = {
+        displayValue: <a href={getAccountActivationUrl(userData.activationKey)} rel="noopener noreferrer" target="_blank" className="word_break">{userData.activationKey}</a>,
+      };
+    } else {
+      dataValue = 'N/A';
+    }
+
+    userAccountData.splice(5, 0,
+      {
+        dataName: 'Activation Key/Link',
+        dataValue,
+      });
+  }
+
   return (
     <section className="mb-3">
       <div className="d-flex flex-row flex-wrap">
         <div className="col-sm-6">
-          <div className="flex-column p-4 m-3 card">
+          <div id="account-table" className="flex-column p-4 m-3 card">
             <h4>Account</h4>
             <Table
               data={userAccountData}
@@ -394,6 +411,7 @@ UserSummary.propTypes = {
     username: PropTypes.string,
     id: PropTypes.number,
     email: PropTypes.string,
+    activationKey: PropTypes.string,
     isActive: PropTypes.bool,
     country: PropTypes.string,
     dateJoined: PropTypes.string,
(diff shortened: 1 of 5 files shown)


src/users/UserSummary.jsx
```diff
@@ -62,6 +62,12 @@ export default function UserSummary({
       dataName: 'Active',
       dataValue: userData.isActive ? 'yes' : 'no',
     },
+    {
+      dataName: 'Activation Key/Link',
+      dataValue: {
+        displayValue: <a href={`${getConfig().LMS_BASE_URL}/activate/${userData.activationKey}`} rel="noopener noreferrer" target="_blank" className="word_break">{userData.activationKey}</a>,
```
awaisdar001: the logic of constructing the URL should be extracted as `getAccountActivationUrl`

awaisdar001 (approved):
just a nit -- are you adding test cases for the new field as well?

azanbinzahid:
> just a nit -- are you adding test cases for the new field as well?

Yes. By adding this filed into our test data, it'll get tested automatically.

src/users/UserSummary.jsx
azanbinzahid: updated

awaisdar001:
> > just a nit -- are you adding test cases for the new field as well?
> 
> Yes. By adding this filed into our test data, it'll get tested automatically.

sweet

Ali-D-Akbar:
Please add a screenshot of the changes.

DawoudSheraz:
> Please add a screenshot of the changes.

+1 on this

src/users/UserSummary.jsx
```diff
@@ -62,6 +62,12 @@ export default function UserSummary({
       dataName: 'Active',
       dataValue: userData.isActive ? 'yes' : 'no',
     },
+    {
+      dataName: 'Activation Key/Link',
+      dataValue: {
+        displayValue: <a href={getAccountActivationUrl(userData.activationKey)} rel="noopener noreferrer" target="_blank" className="word_break">{userData.activationKey}</a>,
+      },
+    },
```
DawoudSheraz: We should verify if this field should be displayed all the time or only when the user is inactive.

src/users/data/test/userSummary.js
```diff
@@ -4,6 +4,7 @@ const UserSummaryData = {
     username: 'edx',
     email: 'edx@example.com',
     id: 1,
+    activationKey: 'db2f834f90fd49afaa3ca2a68a1ae9e1',
```
DawoudSheraz, line 7: verify this data in the unit tests. You will need to add a condition in the existing test.

azanbinzahid:
> Please add a screenshot of the changes.

added

src/users/UserSummary.jsx
azanbinzahid: I have added the condition

src/users/UserSummary.test.jsx
```diff
@@ -26,6 +26,40 @@ describe('User Summary Component Tests', () => {
...
+      expect(tableHeaders).toHaveLength(9);
+    });
+
+    it('Inactive User With No Registration Data ', () => {
+      const InActiveUserData = { ...UserSummaryData.userData, activationKey: null, isActive: false };
+      const InActiveUserSummaryData = { ...UserSummaryData, userData: InActiveUserData };
+      wrapper = mount(<UserSummary {...InActiveUserSummaryData} />);
+      const dataTable = wrapper.find('#account-table table');
+      const tableHeaders = dataTable.find('tbody tr');
+      expect(tableHeaders).toHaveLength(10);
```
DawoudSheraz: Get data from the rows and assert the activation key render's different instances. Asserting on headers is ok but not enough.

azanbinzahid: I have updated tests to get data from row

DawoudSheraz (approved):

PR #120, opened 2021-05-24
fix: search useremail with +

azanbinzahid:
### Description
Fix search user page and URL route to search users with email having plus (+) in them.

Then:
image

Now:
image

### DIFF
--- a/src/users/UserPage.jsx
+++ b/src/users/UserPage.jsx
@@ -1,4 +1,4 @@
-import { camelCaseObject, getConfig, history } from '@edx/frontend-platform';
+import { camelCaseObject, history } from '@edx/frontend-platform';
 import PropTypes from 'prop-types';
 import React, {
   useCallback, useContext, useEffect, useState,
@@ -18,8 +18,19 @@ import UserSummary from './UserSummary';
 
 // Supports urls such as /users/?username={username} and /users/?email={email}
 export default function UserPage({ location }) {
-  const url = getConfig().BASE_URL + location.pathname + location.search;
-  const params = new URL(url).searchParams;
+  // converts query params from url into map e.g. ?param1=value1&param2=value2 -> {param1: value1, param2=value2}
+  const params = new Map(
+    location.search
+      .slice(1) // removes '?' mark from start
+      .split('&')
+      .map(queryParams => queryParams.split('=')),
+  );
+
+  if (params.has('email')) {
+    const email = params.get('email');
+    params.set('email', decodeURIComponent(email));
+  }
+
   const [userIdentifier, setUserIdentifier] = useState(
     params.get('username') || params.get('email') || undefined,
   );
--- a/src/users/data/api.test.js
+++ b/src/users/data/api.test.js
@@ -172,8 +172,18 @@ describe('API', () => {
     });
 
     test.each([successDictResponse, successListResponse])('Successful Fetch by email', async (successResponse) => {
-      mockAdapter.onGet(`${userAccountApiBaseUrl}?email=${testEmail}`).reply(200, successResponse);
+      mockAdapter.onGet(`${userAccountApiBaseUrl}?email=${encodeURIComponent(testEmail)}`).reply(200, successResponse);
       const response = await api.getUser(testEmail);
+      expect(response).toEqual(Array.isArray(successResponse) ? successResponse[0] : successResponse);
+    });
+
+    test.each([successDictResponse, successListResponse])('Successful Fetch by email with +', async (responseType) => {
+      const testEmailWithPlus = 'email+1@example.com';
+      const successResponse = {
+        ...Array.isArray(responseType) ? responseType[0] : responseType, testEmail: testEmailWithPlus,
+      };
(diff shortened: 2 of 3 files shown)


src/users/UserPage.jsx
```diff
@@ -18,8 +18,11 @@ import UserSummary from './UserSummary';
 
 // Supports urls such as /users/?username={username} and /users/?email={email}
 export default function UserPage({ location }) {
-  const url = getConfig().BASE_URL + location.pathname + location.search;
-  const params = new URL(url).searchParams;
+  const params = new Map(location.search.slice(1).split('&').map(kv => kv.split('=')));
+  if (params.has('email')) {
+    params.set('email', decodeURIComponent(params.get('email')));
+  }
+
```
DawoudSheraz: Why is this needed? What happens if decoding is not done? 
Code comment: The code is difficult to understand. Improve variable naming and maybe add an inline comment to explain this special case.

src/users/data/api.test.js
```diff
@@ -8,6 +8,7 @@ import * as api from './api';
 describe('API', () => {
   const testUsername = 'username';
   const testEmail = 'email@example.com';
+  const encodedTestEmail = encodeURIComponent(testEmail);
```
DawoudSheraz: don't introduce a new variable. Use testEmail directly. Change its value to encoded value though.

src/users/data/api.test.js
```diff
@@ -172,11 +173,25 @@ describe('API', () => {
...
 
+    test.each([successDictResponse, successListResponse])('Successful Fetch by email with +', async (successResponseType) => {
+      const testEmailWithPlus = 'email+1@example.com';
+      const encodedTestEmailWithPlus = encodeURIComponent(testEmailWithPlus);
+      let successResponse;
+      if (Array.isArray(successResponseType)) {
+        successResponse = { ...successResponseType[0], testEmail: testEmailWithPlus };
+      } else {
+        successResponse = { ...successResponseType, testEmail: testEmailWithPlus };
+      }
```
DawoudSheraz: you can use the ternary operator to simplify this if condition.

DawoudSheraz (changes requested):
Added a few comments to address.

src/users/UserPage.jsx
azanbinzahid: Earlier method `new URL()` auto converts plus (+) into space without letting us encode/decode only params part. Parsing params from location.search requires string manipulation and then concatenations which makes code much lengthy.  This map converts params in url to key/value pairs. 

Decoding here allows both kind of urls in search i.e. `/?email=abc+d@gmail.com` and `/?email=abc%2Bd%40gmail.com` otherwise only first one will work.

src/users/data/api.test.js
azanbinzahid: using ternary operating for this exceeds line char limit forcing to split lines again.

src/users/data/api.test.js
```diff
@@ -172,11 +173,25 @@ describe('API', () => {
     });
 
     test.each([successDictResponse, successListResponse])('Successful Fetch by email', async (successResponse) => {
-      mockAdapter.onGet(`${userAccountApiBaseUrl}?email=${testEmail}`).reply(200, successResponse);
+      mockAdapter.onGet(`${userAccountApiBaseUrl}?email=${encodedTestEmail}`).reply(200, successResponse);
```
Ali-D-Akbar: Can't we just use the normal value for this? I think it can be written without encoding.

src/users/data/api.test.js
```diff
@@ -172,11 +173,25 @@ describe('API', () => {
...
       const response = await api.getUser(testEmail);
       expect(response).toEqual(Array.isArray(successResponse) ? successResponse[0] : successResponse);
     });
 
+    test.each([successDictResponse, successListResponse])('Successful Fetch by email with +', async (successResponseType) => {
+      const testEmailWithPlus = 'email+1@example.com';
+      const encodedTestEmailWithPlus = encodeURIComponent(testEmailWithPlus);
```
Ali-D-Akbar, line 183: Can be written in a single variable as said before.

src/users/data/api.test.js
azanbinzahid: I tried this, but it fails the test and encoding is needed only when making a get call from mockAdapter. We are now sending encoded email when making a regular api call to server and in this case the expected and real response match.

src/users/data/api.test.js
azanbinzahid:
```
const successResponse = {
  ...Array.isArray(responseType) ? responseType[0] : responseType, testEmail: testEmailWithPlus,
};
```
This could be one way if that's readable?

src/users/UserPage.jsx
azanbinzahid: These are few more approaches for getting query params: https://stackoverflow.com/questions/901115/how-can-i-get-query-string-values-in-javascript

DawoudSheraz (approved):

Ali-D-Akbar (approved):
👍

PR #121, opened 2021-05-24
feat: add view certificate button

DawoudSheraz:
### Description

Add certificate tool with the following features:

- Get certificate details
- Generate certificate
- Regenerate certificate

The backend APIs are the same as old support tools, located at https://github.com/edx/edx-platform/blob/master/lms/djangoapps/certificates/views/support.py. The view cert button is added to the enrollment listing, as requested in the ticket.

image
image
image
image
image

### DIFF
--- a/src/users/data/api.js
+++ b/src/users/data/api.js
@@ -443,3 +443,74 @@ export async function postResetPassword(email) {
   );
   return data;
 }
+
+export async function getCertificate(username, courseKey) {
+  try {
+    const { data } = await getAuthenticatedHttpClient().get(
+      AppUrls.getCertificateUrl(username, courseKey),
+    );
+    return Array.isArray(data) && data.length > 0 ? data[0] : data;
+  } catch (error) {
+    return {
+      errors: [
+        {
+          code: null,
+          dismissible: true,
+          text: error.message,
+          type: 'danger',
+          topic: 'certificates',
+        },
+      ],
+    };
+  }
+}
+
+export async function generateCertificate(username, courseKey) {
+  const formData = new FormData();
+  formData.append('username', username);
+  formData.append('course_key', courseKey);
+  try {
+    const { data } = await getAuthenticatedHttpClient().post(
+      AppUrls.generateCertificateUrl(),
+      formData,
+    );
+    return data;
+  } catch (error) {
+    return {
+      errors: [
+        {
+          code: null,
+          dismissible: true,
+          text: error.message,
+          type: 'danger',
+          topic: 'certificates',
+        },
+      ],
+    };
+  }
+}
+
+export async function regenerateCertificate(username, courseKey) {
+  const formData = new FormData();
+  formData.append('username', username);
+  formData.append('course_key', courseKey);
+  try {
+    const { data } = await getAuthenticatedHttpClient().post(
+      AppUrls.regenerateCertificateUrl(),
+      formData,
+    );
+    return data;
+  } catch (error) {
+    return {
+      errors: [
+        {
+          code: null,
+          dismissible: true,
+          text: error.message,
+          type: 'danger',
+          topic: 'certificates',
+        },
+      ],
+    };
+  }
+}
--- a/src/users/data/api.test.js
+++ b/src/users/data/api.test.js
@@ -5,6 +5,7 @@ import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
 import { enrollmentsData } from './test/enrollments';
 import { downloadableCertificate } from './test/certificates';
 import * as api from './api';
+import * as urls from './urls';
 
 describe('API', () => {
   const testUsername = 'username';
@@ -19,9 +20,9 @@ describe('API', () => {
   const verificationStatusApiUrl = `${getConfig().LMS_BASE_URL}/api/user/v1/accounts/${testUsername}/verification_status/`;
   const licensesApiUrl = `${getConfig().LICENSE_MANAGER_URL}/api/v1/staff_lookup_licenses/`;
   const onboardingStatusApiUrl = `${getConfig().LMS_BASE_URL}/api/edx_proctoring/v1/user_onboarding/status`;
-  const certificatesUrl = `${getConfig().LMS_BASE_URL}/certificates/search?user=${testUsername}&course_id=${testCourseId}`;
-  const generateCertificateUrl = `${getConfig().LMS_BASE_URL}/certificates/generate`;
-  const regenerateCertificateUrl = `${getConfig().LMS_BASE_URL}/certificates/regenerate`;
+  const certificatesUrl = urls.getCertificateUrl(testUsername, testCourseId);
+  const generateCertificateUrl = urls.generateCertificateUrl();
+  const regenerateCertificateUrl = urls.regenerateCertificateUrl();
 
   let mockAdapter;
 
--- a/src/users/data/api.test.js
+++ b/src/users/data/api.test.js
@@ -3,11 +3,13 @@ import { getConfig } from '@edx/frontend-platform';
 import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
 
 import { enrollmentsData } from './test/enrollments';
+import { downloadableCertificate } from './test/certificates';
 import * as api from './api';
 
 describe('API', () => {
   const testUsername = 'username';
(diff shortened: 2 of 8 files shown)


src/users/data/api.test.js
```diff
@@ -17,6 +19,9 @@ describe('API', () => {
   const verificationStatusApiUrl = `${getConfig().LMS_BASE_URL}/api/user/v1/accounts/${testUsername}/verification_status/`;
   const licensesApiUrl = `${getConfig().LICENSE_MANAGER_URL}/api/v1/staff_lookup_licenses/`;
   const onboardingStatusApiUrl = `${getConfig().LMS_BASE_URL}/api/edx_proctoring/v1/user_onboarding/status`;
+  const certificatesUrl = `${getConfig().LMS_BASE_URL}/certificates/search?user=${testUsername}&course_id=${testCourseId}`;
+  const generateCertificateUrl = `${getConfig().LMS_BASE_URL}/certificates/generate`;
+  const regenerateCertificateUrl = `${getConfig().LMS_BASE_URL}/certificates/regenerate`;
```
Ali-D-Akbar: Why're you not using getCertificateUrl, regenerateCertificateUrl and regenerateCertificateUrl from urls.js?

src/users/enrollments/Certificates.test.jsx
```diff
@@ -0,0 +1,237 @@
...
+describe('Certificate component', () => {
+  let apiMock; let wrapper;
+  const testUser = 'testUser';
+  const testCourseId = 'course-v1:testX+test123+2030';
+
+  const props = {
+    username: testUser,
+    courseId: testCourseId,
+    closeHandler: jest.fn(() => {}),
+  };
```
Ali-D-Akbar, line 25: Move this in src/users/data/test/certificates.js.

src/users/enrollments/Certificates.test.jsx
```diff
@@ -0,0 +1,237 @@
...
+
+      expect(generateButton.prop('disabled')).toBeTruthy();
+      expect(generateApiMock).toHaveBeenCalledTimes(1);
+      expect(wrapper.find('h3').text()).toEqual('Status: Generating New Certificate ');
+
+      // Once the generation api call is successful, the status text and button will revert.
+      await waitForComponentToPaint(wrapper);
+      generateButton = wrapper.find('button#generate-certificate');
+      expect(wrapper.find('h3').length).toEqual(0);
+      expect(generateButton.prop('disabled')).toBeFalsy();
```
Ali-D-Akbar: why is the generate-certificate button still not disabled after flow has been successful?
And are we showing some success message on success of certificate generation?

Ali-D-Akbar, line 145: why is the generate-certificate button still not disabled after flow has been successful?
And are we showing some success message on success of certificate generation?

Ali-D-Akbar (changes requested):
A handful of queries to be addressed:

1. Can you please add a screenshot for regenerate button scenario? Just wanted to confirm that only one of the button is shown
2. What does a user see as a verification that an operation was successful? Some sort of a success message or the whole page/component refresh?

src/users/enrollments/Certificates.test.jsx
DawoudSheraz: This is not the exact data. The data is already in cert.js.

src/users/enrollments/Certificates.test.jsx
DawoudSheraz: Because it is still possible to generate/regenerate after one attempt. The backend does not restrict reattempts for generation/regeneration. Ideally, the generate button would change into the regenerate button. But the mock in this current test is returning the generatable certificate data.

src/users/data/api.test.js
DawoudSheraz: Right, I will change it

DawoudSheraz:
> A handful of queries to be addressed:
> 
> 1. Can you please add a screenshot for regenerate button scenario? Just wanted to confirm that only one of the button is shown
> 2. What does a user see as a verification that an operation was successful? Some sort of a success message or the whole page/component refresh?

1. Ok, although there won't be much difference. Only status text is changed.
2. If the generate/regenerate is successful, only the cert component reloads. This behavior is the same as the old support tools, where the reloading is done via django.

src/users/data/api.js
```diff
@@ -443,3 +443,74 @@ export async function postResetPassword(email) {
   );
   return data;
 }
+
+export async function getCertificate(username, courseKey) {
+  try {
+    const { data } = await getAuthenticatedHttpClient().get(
+      AppUrls.getCertificateUrl(username, courseKey),
+    );
+    return Array.isArray(data) && data.length > 0 ? data[0] : data;
```
azanbinzahid, line 452: In what cases this endpoint will return an array or single item?

src/users/data/api.js
```diff
@@ -443,3 +443,74 @@ export async function postResetPassword(email) {
...
+export async function generateCertificate(username, courseKey) {
+  const formData = new FormData();
+  formData.append('username', username);
+  formData.append('course_key', courseKey);
+  try {
+    const { data } = await getAuthenticatedHttpClient().post(
+      AppUrls.generateCertificateUrl(),
+      formData,
+    );
+    return data;
```
azanbinzahid, line 477: What does `data` contain in this case? from [here](https://github.com/edx/edx-platform/blob/master/lms/djangoapps/certificates/views/support.py#L223) it says its a http response. Are we using it anywhere?

src/users/data/urls.js
```diff
@@ -69,3 +69,15 @@ export const getAccountActivationUrl = (activationKey) => `${
 export const getOnboardingStatusUrl = (courseId, username) => `${
   LMS_BASE_URL
 }/api/edx_proctoring/v1/user_onboarding/status?course_id=${encodeURIComponent(courseId)}&username=${encodeURIComponent(username)}`;
+
+export const getCertificateUrl = (username, courseKey) => `${
+  LMS_BASE_URL
+}/certificates/search?user=${username}&course_id=${courseKey}`;
```
azanbinzahid, line 75: we may need to encode `courseKey`.

Ali-D-Akbar (approved):

src/users/enrollments/Certificates.jsx
```diff
@@ -0,0 +1,198 @@
...
+
+      {!certificate && !displayCertErrors && <PageLoading srMessage="Loading" /> }
+      {displayCertErrors && <AlertList topic="certificates" />}
+
+      {certificate && !displayCertErrors && (
+
+      <div className="m-3">
+        <h2>Course ID: {certificate.courseKey}</h2>
+        {status && (<h3>Status: {status} </h3>)}
+        <table className="table">
```
azanbinzahid, line 122: Is there a specific reason for not using `Table` component?

src/users/data/api.js
DawoudSheraz: Yeah, it is just a response. We are not using it anywhere though. The return value is used only in tests to verify the api behavior.

src/users/enrollments/Certificates.jsx
DawoudSheraz: Yes, the Table component is more suited for listing with headings on top and data listed vertically. In cert case, the headers are on the left and data on right i.e. one header and one piece of data are on the same row. furthermore, the Table component has many other operations not needed in this case.

src/users/data/api.js
DawoudSheraz: If course id is mentioned, then the single item, otherwise a list of certificates. We might not need this check. I added it for cautious purposes(unexpected data on stage/prod). Once the changes are live, I will validate and remove the unnecessary check.

src/users/data/urls.js
DawoudSheraz: Hmm, the backend is running quote_plus on the course key. https://github.com/edx/edx-platform/blob/master/lms/djangoapps/certificates/views/support.py#L109. I will check if the encoding is necessary.
https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote_plus

DawoudSheraz: Encoding is not needed. Encoding the courseId results in 404 on backend.

azanbinzahid (approved):

PR #126, opened 2021-06-07
feat: show onboarding status for user

azanbinzahid:
## Description
This PR adds `Proctoring Information` table which fetch user onboarding data against this [endpoint](https://github.com/edx/edx-proctoring/blob/master/edx_proctoring/views.py#L444) which is a wrapper of `Verificient` service. 

## Linked Issue Ticket
[PROD-2358](https://openedx.atlassian.net/browse/PROD-2358) contains detailed discussion and decisions. 

## Local Dependency
[edx/mockprock](https://github.com/edx/edx-proctoring/blob/master/docs/developing.rst#using-mockprock-as-a-backend) should be setup and running in devstack.

## Testing
- Edge cases needs to be tested on stage
- Feedback from support is must 

## Rebase and Commit Squash
May need to rebase with master and commits will be squashed into one with `feat: show onboarding status for user` message before merge. 

## Screenshot
image
image

### DIFF
--- a/src/users/UserSummary.jsx
+++ b/src/users/UserSummary.jsx
@@ -1,15 +1,17 @@
 import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import { Modal, Button, Input } from '@edx/paragon';
+import { getConfig } from '@edx/frontend-platform';
 import { postTogglePasswordStatus, postResetPassword } from './data/api';
 import Table from '../Table';
-import { formatDate } from '../utils';
+import { formatDate, titleCase } from '../utils';
 import { getAccountActivationUrl } from './data/urls';
 
 export default function UserSummary({
   userData,
   verificationData,
   ssoRecords,
+  onboardingData,
   changeHandler,
 }) {
   const [ssoModalIsOpen, setSsoModalIsOpen] = useState(false);
@@ -151,6 +153,7 @@ export default function UserSummary({
       key: 'extra',
     },
   ];
+
   const userPasswordHistoryColumns = [
     {
       label: 'Date',
@@ -241,6 +244,35 @@ export default function UserSummary({
     },
   }));
 
+  const proctoringData = [onboardingData].map(result => ({
+    status: result.onboardingStatus ? titleCase(result.onboardingStatus) : 'Not Started',
+    expirationDate: formatDate(result.expirationDate),
+    onboardingReleaseDate: formatDate(result.onboardingReleaseDate),
+    onboardingLink: result.onboardingLink ? {
+      displayValue: <a href={`${getConfig().LMS_BASE_URL}${result.onboardingLink}`} rel="noopener noreferrer" target="_blank" className="word_break">Link</a>,
+      value: result.onboardingLink,
+    } : 'N/A',
+  }));
+
+  const proctoringColumns = [
+    {
+      label: 'Onboarding Status',
+      key: 'status',
+    },
+    {
+      label: 'Expiration Date',
+      key: 'expirationDate',
+    },
+    {
+      label: 'Release Date',
+      key: 'onboardingReleaseDate',
+    },
+    {
+      label: 'Onboarding Link',
+      key: 'onboardingLink',
+    },
+  ];
+
   if (!userData.isActive) {
     let dataValue;
     if (userData.activationKey !== null) {
@@ -304,6 +336,14 @@ export default function UserSummary({
                 id="idv-data"
                 data={IdvData}
                 columns={idvColumns}
+              />
+            </div>
+            <div className="flex-column p-4 m-3 card">
+              <h4>Proctoring Information</h4>
+              <Table
+                id="proctoring-data"
+                data={proctoringData}
+                columns={proctoringColumns}
               />
             </div>
             <div className="flex-column p-4 m-3 card">
@@ -428,6 +468,12 @@ UserSummary.propTypes = {
     extraData: PropTypes.shape([]),
   }),
   ssoRecords: PropTypes.shape([]),
+  onboardingData: PropTypes.shape({
+    onboardingStatus: PropTypes.string,
+    expirationDate: PropTypes.string,
+    onboardingLink: PropTypes.string,
+    onboardingReleaseDate: PropTypes.string,
+  }),
   changeHandler: PropTypes.func.isRequired,
 };
 
@@ -435,4 +481,5 @@ UserSummary.defaultProps = {
   userData: null,
   verificationData: null,
   ssoRecords: [],
+  onboardingData: null,
 };
--- a/src/users/UserSummary.test.jsx
+++ b/src/users/UserSummary.test.jsx
@@ -4,6 +4,7 @@ import React from 'react';
 import * as api from './data/api';
 import UserSummary from './UserSummary';
 import UserSummaryData from './data/test/userSummary';
+import { formatDate, titleCase } from '../utils';
 
 const getActivationKeyRow = (data) => {
   const wrapper = mount(<UserSummary {...data} />);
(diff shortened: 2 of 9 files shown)


src/users/UserSummary.test.jsx
```diff
@@ -60,6 +61,36 @@ describe('User Summary Component Tests', () => {
...
+
+      expect(ComponentOnboardingData.onboardingStatus).toEqual(ExpectedOnboardingData.onboardingStatus);
+      expect(ComponentOnboardingData.expirationDate).toEqual(ExpectedOnboardingData.expirationDate);
+      expect(ComponentOnboardingData.onboardingReleaseDate).toEqual(ExpectedOnboardingData.onboardingReleaseDate);
+      expect(ComponentOnboardingData.onboardingLink).toEqual(ExpectedOnboardingData.onboardingLink);
+    });
+  });
+
+  describe('Proctoring Data', () => {
+    it('Onboarding Status', () => {
```
Ali-D-Akbar, line 76: I think it'd be `'Onboarding Data'`

src/users/data/api.test.js
```diff
@@ -36,6 +37,46 @@ describe('API', () => {
...
+    });
+
+    it('No Record for Onboaring Status Fetch ', async () => {
+      mockAdapter.onGet(url).reply(() => throwError(404, ''));
+
+      const response = await api.getOnboardingStatus(data, testUsername);
+      expect(response).toEqual({ ...expectedSuccessResponse, onboardingStatus: 'No Record Found' });
+    });
+
+    it('Unexpected error', async () => {
```
Ali-D-Akbar, line 70: Why do we need to check Unexpected error in this case?

src/users/UserSummary.test.jsx
```diff
@@ -60,6 +61,36 @@ describe('User Summary Component Tests', () => {
...
+      const dataTable = wrapper.find('Table#proctoring-data');
+      const dataBody = dataTable.find('tbody tr td');
+      expect(dataBody).toHaveLength(4);
+      expect(dataBody.at(0).text()).toEqual(titleCase(UserSummaryData.onboardingData.onboardingStatus));
+      expect(dataBody.at(3).text()).toEqual('Link');
```
Ali-D-Akbar: Check for other 2 values as well.

Ali-D-Akbar (commented):
Some comments added to be addressed.

src/users/UserSummary.jsx
```diff
@@ -428,11 +468,20 @@ UserSummary.propTypes = {
     extraData: PropTypes.shape([]),
   }),
   ssoRecords: PropTypes.shape([]),
+  onboardingData: PropTypes.shape({
+    onboardingStatus: PropTypes.string,
+    expirationDate: PropTypes.string,
+    onboardingLink: PropTypes.string,
+    onboardingPastDue: PropTypes.bool,
+    onboardingReleaseDate: PropTypes.string,
+    reviewRequirementsUrl: PropTypes.string,
```
DawoudSheraz: Do not add the props if they are not being used.

src/users/UserSummary.test.jsx
DawoudSheraz: or Onboarding Status Data

src/users/UserSummary.test.jsx
DawoudSheraz: Yes, check all values.

src/users/UserSummary.test.jsx
```diff
@@ -60,6 +61,36 @@ describe('User Summary Component Tests', () => {
...
+    });
+
+    it('Null Onboarding Status', () => {
```
DawoudSheraz: No Onboarding status data/Missing onboarding status data

src/users/UserSummary.test.jsx
```diff
@@ -60,6 +61,36 @@ describe('User Summary Component Tests', () => {
...
+      const onboardingData = { ...UserSummaryData.onboardingData, onboardingStatus: null, onboardingLink: null };
+      const userData = { ...UserSummaryData, onboardingData };
+      wrapper = mount(<UserSummary {...userData} />);
+      const dataTable = wrapper.find('Table#proctoring-data');
+      const dataBody = dataTable.find('tbody tr td');
+      expect(dataBody).toHaveLength(4);
+      expect(dataBody.at(0).text()).toEqual('Not Started');
+      expect(dataBody.at(3).text()).toEqual('N/A');
```
DawoudSheraz: validate all fields here as well

src/users/data/api.js
```diff
@@ -175,6 +175,45 @@ export async function getLicense(userEmail) {
...
+  paidEnrollments.sort((a, b) => (a.created < b.created));
+
+  if (paidEnrollments.length === 0) {
+    return {
+      ...defaultResponse,
+      onboardingStatus: 'No Paid Enrollment',
+    };
+  }
+
+  const courseId = paidEnrollments[0].course_id;
```
DawoudSheraz, line 199: Isn't API response transformed into camelcase? Shouldn't course_id be courseId?

src/users/data/api.js
```diff
@@ -175,6 +175,45 @@ export async function getLicense(userEmail) {
...
+  try {
+    const { data } = await getAuthenticatedHttpClient().get(
+      AppUrls.getOnboardingStatusUrl(courseId, username),
+    );
+    return data;
+  } catch (error) {
+    const { httpErrorStatus } = error && error.customAttributes;
```
DawoudSheraz: are we certain there will be custom attributes in the error?

src/users/data/api.test.js
```diff
@@ -36,6 +37,46 @@ describe('API', () => {
...
+      onboardingPastDue: null,
+      onboardingReleaseDate: null,
+      reviewRequirementsUrl: null,
+    };
+
+    // prepare enrollments data
+    const { data } = enrollmentsData;
+    data[1].mode = 'verified';
+    data[1].course_id = data[1].courseId;
+    const url = `${onboardingStatusApiUrl}?course_id=${encodeURIComponent(data[1].courseId)}&username=${encodeURIComponent(testUsername)}`;
```
DawoudSheraz, line 54: why encodeUrlComponent is needed here? Does it work without it? Why/Why not?

src/users/data/api.js
```diff
@@ -175,6 +175,45 @@ export async function getLicense(userEmail) {
...
+    if (httpErrorStatus === 404) {
+      return {
+        ...defaultResponse,
+        onboardingStatus: 'No Record Found',
+      };
+    }
+    throw error;
```
DawoudSheraz: return error instead of throwing error. You can use alert error format.

src/users/data/urls.js
```diff
@@ -69,3 +69,7 @@ export const getResetPasswordUrl = () => `${
 export const getAccountActivationUrl = (activationKey) => `${
   LMS_BASE_URL
 }/activate/${activationKey}`;
+
+export const getOnboardingStatusUrl = (courseId, username) => `${
+  LMS_BASE_URL
+}/api/edx_proctoring/v1/user_onboarding/status?course_id=${encodeURIComponent(courseId)}&username=${encodeURIComponent(username)}`;
```
DawoudSheraz, line 71: why is encoding needed?

src/utils/index.js
```diff
@@ -26,3 +26,7 @@ export function sort(firstElement, secondElement, key, direction) {
   }
   return 0;
 }
+
+export function titleCase(str) {
+  return str.toLowerCase().replace(/_/g, ' ').replace(/\b(\w)/g, s => s.toUpperCase());
+}
```
DawoudSheraz, line 33: add a small jsdoc explaining what this function is doing. Regex tend to be difficult to interpret, so having a one-liner would make it easier for devs to understand what this function is doing.

src/users/data/api.test.js
azanbinzahid: In case some error code comes for which we don't have explicit message.

src/users/data/urls.js
azanbinzahid: Since course name has non-alpha numeric characters especially `+` sign therefor encoding is needed. See similar call [here](https://github.com/edx/frontend-app-learning/blob/73fa56d40105d7c15022303bfe2afef25fce68fe/src/course-home/data/api.js#L146)

src/users/data/api.test.js
azanbinzahid: Same [reason](https://github.com/edx/frontend-app-support-tools/pull/126#discussion_r653392012)

src/users/data/api.test.js
Ali-D-Akbar: I think this test doesn't really do anything. This can be removed.

src/utils/index.js
azanbinzahid: updated

src/users/data/api.js
azanbinzahid: updated with an if condition

src/users/data/api.test.js
azanbinzahid: it checks for the throw Error condition like what if there is no 404 error.

src/users/data/api.js
DawoudSheraz: @azanbinzahid To keep the majority consistency, please return the error instead of throwing. Use error format used by alert component.

src/users/data/api.js
```diff
@@ -199,6 +238,7 @@ export async function getAllUserData(userIdentifier) {
     ssoRecords = await getSsoRecords(user.username);
     user.passwordStatus = await getUserPasswordStatus(user.username);
     licenses = await getLicense(user.email);
+    onboardingStatus = await getOnboardingStatus(enrollments, user.username);
```
DawoudSheraz, line 244: If the error is thrown, you would need a catch statement here. To avoid the extra bits of code, return the error and let the component deal with the rendering.

src/utils/index.js
```diff
@@ -26,3 +26,8 @@ export function sort(firstElement, secondElement, key, direction) {
   }
   return 0;
 }
+
+/** Convert a string containing space and/or underscore (snake_case) into titleCase. */
```
DawoudSheraz: nit: add an example

DawoudSheraz (commented):
a few nits to address

src/users/data/api.js
azanbinzahid: I have updated the implementation, please review.

src/users/data/api.test.js
```diff
@@ -36,6 +37,44 @@ describe('API', () => {
...
+
+    it('Successfull Fetch ', async () => {
```
DawoudSheraz: nit: successful

DawoudSheraz (approved):
one small nit to cater before merging

src/users/data/api.js
azanbinzahid: It gets converted to camelCase later on in [UserPage](https://github.com/edx/frontend-app-support-tools/blob/ca3d25bc2755e465d2d04cf580bd34106998d824/src/users/UserPage.jsx#L93) component.

src/users/data/api.test.js
Ali-D-Akbar: If we're not throwing any errors, we don't need to check it anymore. Please remove this test.

Ali-D-Akbar (approved):
Since the throw error approach has been removed, I think that the `Unexpected error` test in api.test.js has no purpose anymore.

The rest looks good.

PR #127, opened 2021-06-23
fix: sort order for enrollments

azanbinzahid:
Updates on PR #126

- Fix sort order for enrollments ([see](https://www.javascripttutorial.net/javascript-array-sort/) how array.sort() works, dates example in particular)
- Remove release date field

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -31,3 +31,10 @@ export function sort(firstElement, secondElement, key, direction) {
 export function titleCase(str) {
   return str.toLowerCase().replace(/_/g, ' ').replace(/\b(\w)/g, s => s.toUpperCase());
 }
+
+/**  Compare dates function for array.sort() */
+export function sortedCompareDates(x, y, asc) {
+  const a = new Date(x);
+  const b = new Date(y);
+  return asc ? a - b : b - a;
+}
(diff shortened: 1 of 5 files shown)


src/utils/index.js
```diff
@@ -31,3 +31,10 @@ export function sort(firstElement, secondElement, key, direction) {
 export function titleCase(str) {
   return str.toLowerCase().replace(/_/g, ' ').replace(/\b(\w)/g, s => s.toUpperCase());
 }
+
+/**  Compare dates function for array.sort() */
+export function compareDates(x, y, asc) {
+  const a = new Date(x);
+  const b = new Date(y);
+  return asc ? a - b : b - a;
+}
```
DawoudSheraz: the function name should be sortedCompareDates because the code is changing the positions of the dates based on the sort param. It isn't only doing the comparison. It is changing the decision of comparison by relating it to sort

azanbinzahid: I believe this function is only doing comparison, if we see its return type it will be 0, greater than 0 or less than 0, based on which `array.sort()` itself changes position of elements. Refer to the convention and explanation [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort).

DawoudSheraz: Yes, but the way comparison is done is impacted by the argument. The return value and the sorting are affected by that extra argument.

azanbinzahid: Just updated the function name.

DawoudSheraz (approved):

PR #132, opened 2021-06-29
refactor: self-contain license component

azanbinzahid:
### Description
This PR refactors license component towards self-containment, essentially moving its API calls into its own component. This  change is a part of optimization for support tools [PROD-2363](https://openedx.atlassian.net/browse/PROD-2363). 

### Linked Ticket:
[PROD-2394](https://openedx.atlassian.net/browse/PROD-2394)

### DIFF
--- a/src/users/licenses/Licenses.jsx
+++ b/src/users/licenses/Licenses.jsx
@@ -2,23 +2,37 @@ import React, {
   useMemo,
   useState,
   useCallback,
+  useEffect,
 } from 'react';
 import PropTypes from 'prop-types';
 import { Collapsible, Badge } from '@edx/paragon';
+import { camelCaseObject } from '@edx/frontend-platform';
 import Table from '../../Table';
 import { formatDate, sort } from '../../utils';
+import { getLicense } from '../data/api';
 
 export default function Licenses({
-  data, status, expanded,
+  userEmail, expanded,
 }) {
   const [sortColumn, setSortColumn] = useState('status');
   const [sortDirection, setSortDirection] = useState('desc');
+  const [licenses, setLicensesData] = useState([]);
+  const [status, setStatus] = useState('Loading...');
+
+  useEffect(() => {
+    getLicense(userEmail).then((data) => {
+      const camelCaseData = camelCaseObject(data);
+      setLicensesData(camelCaseData.results);
+      setStatus(camelCaseData.status);
+    });
+  }, [userEmail]);
+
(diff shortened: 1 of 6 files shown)


src/users/licenses/Licenses.jsx
```diff
@@ -2,23 +2,37 @@ import React, {
...
+  const [status, setStatus] = useState('');
+
+  useEffect(() => {
+    getLicense(userEmail).then((data) => {
+      const camelCaseData = camelCaseObject(data);
+      setLicensesData(camelCaseData.results);
+      setStatus(camelCaseData.status);
+    });
+  }, [userEmail]);
+
```
DawoudSheraz, line 29: What will be the behavior/render before the API call is successful? Empty table? or loading bar?

DawoudSheraz (commented):
A comment/question to address.

src/users/licenses/Licenses.test.jsx
```diff
@@ -12,22 +13,35 @@ const LicensesPageWrapper = (props) => (
...
+  let apiMock; let wrapper;
+  const props = {
+    userEmail: 'test@example.com',
+    expanded: true,
+  };
 
-  beforeEach(() => {
-    wrapper = mount(<LicensesPageWrapper {...licensesData} />);
+  afterEach(() => {
+    apiMock.mockRestore();
```
Ali-D-Akbar: We don't need to `mockRestore` if we're using `mockImplementationOnce`.
Also, I'd suggest `wrapper.unmount();` in afterEach so that failing one test doesn't affect the others in any way.

Ali-D-Akbar (changes requested):
Just a small change in tests. 
I'd also suggest adding a loading bar while the component fetches the data from API.

src/users/licenses/Licenses.jsx
azanbinzahid: Empty table since initial state of license is empty list. Didn't implement loading bar since its collapsible but should we?

src/users/licenses/Licenses.test.jsx
azanbinzahid: thanks, let me look at it.

azanbinzahid: updated as suggested.

Ali-D-Akbar (approved):
LGTM.

DawoudSheraz (approved):
:shipit:

PR #133, opened 2021-06-30
refactor: move api calls in entitlement component

Ali-D-Akbar:
[PROD-2395](https://openedx.atlassian.net/browse/PROD-2395)
Entitlement api calls have been moved from UserPage.jsx to Entitlements.jsx hence the component no longer depends on UserPage.jsx.

There are two visual changes in this PR:
1. I've also added a Page Loading while the entitlement is fetching data from api. Also the tests have been thoroughly changed according to the entitlement changes: (Edit: the loader has been moved inside the component)
image

2. Also added error handling as shown below: 
Screenshot 2021-07-01 at 2 24 06 PM

### DIFF
--- a/src/users/UserPage.jsx
+++ b/src/users/UserPage.jsx
@@ -172,7 +172,6 @@ export default function UserPage({ location }) {
           />
           <Entitlements
             user={data.user.username}
-            data={data.entitlements}
             changeHandler={handleEntitlementsChange}
             expanded={showEntitlements}
           />
--- a/src/users/data/api.js
+++ b/src/users/data/api.js
@@ -5,17 +5,31 @@ import * as AppUrls from './urls';
 import { isEmail, sortedCompareDates } from '../../utils';
 
 export async function getEntitlements(username, page = 1) {
-  const baseURL = AppUrls.getEntitlementUrl();
-  const queryString = `user=${username}&page=${page}`;
-  const { data } = await getAuthenticatedHttpClient().get(
-    `${baseURL}?${queryString}`,
-  );
-  if (data.next !== null) {
-    const nextPageData = await getEntitlements(username, data.current_page + 1);
-    data.results = data.results.concat(nextPageData.results);
+  try {
+    const baseURL = AppUrls.getEntitlementUrl();
+    const queryString = `user=${username}&page=${page}`;
+    const { data } = await getAuthenticatedHttpClient().get(
+      `${baseURL}?${queryString}`,
+    );
+    if (data.next !== null) {
+      const nextPageData = await getEntitlements(username, data.current_page + 1);
+      data.results = data.results.concat(nextPageData.results);
+      return data;
+    }
     return data;
+  } catch (error) {
+    return {
+      errors: [
+        {
+          code: null,
+          dismissible: true,
+          text: JSON.parse(error.customAttributes.httpErrorResponseData),
+          type: 'danger',
+          topic: 'entitlements',
+        },
+      ],
+    };
   }
-  return data;
 }
 
 export async function getEnrollments(username) {
@@ -221,7 +235,6 @@ export async function getOnboardingStatus(enrollments, username) {
 export async function getAllUserData(userIdentifier) {
   const errors = [];
   let user = null;
-  let entitlements = [];
   let enrollments = [];
   let verificationStatus = null;
   let ssoRecords = null;
@@ -236,7 +249,6 @@ export async function getAllUserData(userIdentifier) {
     }
   }
   if (user !== null) {
-    entitlements = await getEntitlements(user.username);
     enrollments = await getEnrollments(user.username);
     verificationStatus = await getUserVerificationStatus(user.username);
     ssoRecords = await getSsoRecords(user.username);
@@ -247,7 +259,6 @@ export async function getAllUserData(userIdentifier) {
   return {
     errors,
     user,
-    entitlements,
     enrollments,
     verificationStatus,
     ssoRecords,
--- a/src/users/data/api.test.js
+++ b/src/users/data/api.test.js
@@ -348,7 +348,6 @@ describe('API', () => {
 
     it('Successful User Data Retrieval', async () => {
       mockAdapter.onGet(`${userAccountApiBaseUrl}/${testUsername}`).reply(200, successDictResponse);
-      mockAdapter.onGet(`${entitlementsApiBaseUrl}&page=1`).reply(200, { results: [], next: null });
       mockAdapter.onGet(enrollmentsApiUrl).reply(200, []);
       mockAdapter.onGet(ssoRecordsApiUrl).reply(200, []);
       mockAdapter.onGet(verificationDetailsApiUrl).reply(200, {});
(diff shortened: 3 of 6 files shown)


src/users/data/api.js
```diff
@@ -10,7 +10,7 @@ export async function getEntitlements(username, page = 1) {
   const { data } = await getAuthenticatedHttpClient().get(
     `${baseURL}?${queryString}`,
   );
-  if (data.next !== null) {
+  if (data.next && data.next !== null) {
```
DawoudSheraz: why is there a change in the condition needed?

src/users/entitlements/Entitlements.jsx
```diff
@@ -26,12 +27,20 @@ export default function Entitlements({
...
   const [entitlementDetailModalIsOpen, setEntitlementDetailModalIsOpen] = useState(false);
   const [entitlementSupportDetailsTitle, setEntitlementSupportDetailsTitle] = useState('');
   const [entitlementSupportDetails, setEntitlementSupportDetails] = useState([]);
   const formRef = useRef(null);
   const summaryRef = useRef(null);
 
+  useEffect(() => {
+    getEntitlements(user).then((result) => {
+      const camelCaseResult = camelCaseObject(result);
+      setEntitlementData(camelCaseResult);
```
DawoudSheraz: what if there is an error during the fetch? How will that be dealt with?

src/users/data/api.js
Ali-D-Akbar: Oh, this is unnecessary. Removing it.

src/users/UserPage.jsx
```diff
@@ -173,7 +173,6 @@ export default function UserPage({ location }) {
           />
           <Entitlements
             user={data.user.username}
-            data={data.entitlements}
             changeHandler={handleEntitlementsChange}
```
azanbinzahid, line 175: Why can't we move this handler into its own component?

src/users/entitlements/Entitlements.jsx
```diff
@@ -296,32 +305,31 @@ export default function Entitlements({
...
-          data={tableDataSortable.sort(
-            (firstElement, secondElement) => sort(firstElement, secondElement, sortColumn, sortDirection),
-          )}
-          columns={columns}
-          tableSortable
-          defaultSortedColumn="created"
-          defaultSortDirection="desc"
-        />
-      </Collapsible>
+      {!entitlementData && <PageLoading srMessage="Loading" />}
```
azanbinzahid: this can be improved, like adding information of what is being loaded. Ideally the loader should be inside table body so that it does not distort UI before/after loading.

src/users/entitlements/Entitlements.jsx
```diff
@@ -296,32 +305,31 @@ export default function Entitlements({
...
+      {entitlementData && (
```
azanbinzahid: it can be a proper if condition combining above line like if there is data then table (or table body) otherwise loader.

src/users/entitlements/Entitlements.test.jsx
```diff
@@ -16,9 +16,16 @@ const EntitlementsPageWrapper = (props) => (
 
 describe('Entitlements Listing', () => {
   let wrapper;
-
-  beforeEach(() => {
-    wrapper = mount(<EntitlementsPageWrapper {...entitlementsData} />);
+  const entitlementWrapperData = {
```
azanbinzahid: can be renamed to props (optional)

azanbinzahid (changes requested):
Some changes to address ^

src/users/UserPage.jsx
Ali-D-Akbar: The change in Entitlement is subjected to the states of UserPage.jsx as it uses these handlers to expand/hide some collapsible components. In this case, the handler will expand entitlement collapsible and hide the rest.

src/users/entitlements/Entitlements.jsx
Ali-D-Akbar: I'd still go for a loading bar below "Create New Enrollment" eventually serves the same role as a frequent user will understand that there will be a list of entitlements once it has been loaded.

src/users/entitlements/Entitlements.test.jsx
Ali-D-Akbar: Done.

src/users/entitlements/Entitlements.jsx
Ali-D-Akbar: Changed this with a ternary operator.

src/users/entitlements/Entitlements.jsx
```diff
@@ -296,32 +312,30 @@ export default function Entitlements({
...
+            data={tableDataSortable.sort(
+              (firstElement, secondElement) => sort(firstElement, secondElement, sortColumn, sortDirection),
+            )}
+            columns={columns}
+            tableSortable
+            defaultSortedColumn="created"
+            defaultSortDirection="desc"
+          />
+        </Collapsible>
+      )}
```
DawoudSheraz, line 328: We expect the errors to be displayed via alert list in entitlments as well, even though we are not doing it right now. This ternary operator will make it difficult to adjust the third code segment.

src/users/data/api.test.js
```diff
@@ -348,7 +348,6 @@ describe('API', () => {
 
     it('Successful User Data Retrieval', async () => {
       mockAdapter.onGet(`${userAccountApiBaseUrl}/${testUsername}`).reply(200, successDictResponse);
-      mockAdapter.onGet(`${entitlementsApiBaseUrl}&page=1`).reply(200, { results: [], next: null });
```
DawoudSheraz, line 351: why was this removed?

src/users/entitlements/Entitlements.jsx
```diff
@@ -155,7 +170,7 @@ export default function Entitlements({
               id="reissue"
               type="button"
               variant="outline-primary mt-2 mr-2"
-              disabled={!entitlement.enrollmentCourseRun}
+              disabled={Boolean(!entitlement.enrollmentCourseRun)}
```
DawoudSheraz, line 173: why is Boolean call needed?

src/users/entitlements/Entitlements.jsx
```diff
@@ -26,12 +28,25 @@ export default function Entitlements({
...
   const [entitlementSupportDetails, setEntitlementSupportDetails] = useState([]);
   const formRef = useRef(null);
   const summaryRef = useRef(null);
 
+  useEffect(() => {
+    getEntitlements(user).then((result) => {
+      const camelCaseResult = camelCaseObject(result);
+      if (camelCaseResult.errors) {
+        camelCaseResult.errors.forEach(error => add(error));
+        setEntitlementData({ results: [] });
```
DawoudSheraz, line 43: @Ali-D-Akbar Please cater for this missed coverage. That is why coverage is dropping.

DawoudSheraz (approved):

azanbinzahid (approved):

PR #139, opened 2021-07-09
feat: add error handling in enrollment and onboardingStatus component

Ali-D-Akbar:
[PROD-2419](https://openedx.atlassian.net/browse/PROD-2419)

Screenshot 2021-07-09 at 4 43 07 PM

Screenshot 2021-07-09 at 5 49 22 PM

### DIFF
--- a/src/users/data/api.js
+++ b/src/users/data/api.js
@@ -33,10 +33,24 @@ export async function getEntitlements(username, page = 1) {
 }
 
 export async function getEnrollments(username) {
-  const { data } = await getAuthenticatedHttpClient().get(
-    AppUrls.getEnrollmentsUrl(username),
-  );
-  return data;
+  try {
+    const { data } = await getAuthenticatedHttpClient().get(
+      AppUrls.getEnrollmentsUrl(username),
+    );
+    return data;
+  } catch (error) {
+    return {
+      errors: [
+        {
+          code: null,
+          dismissible: true,
+          text: JSON.parse(error.customAttributes.httpErrorResponseData),
+          type: 'danger',
+          topic: 'enrollments',
+        },
+      ],
+    };
+  }
 }
 
 export async function getSsoRecords(username) {
@@ -217,6 +231,13 @@ export async function getOnboardingStatus(enrollments, username) {
     onboardingReleaseDate: null,
     reviewRequirementsUrl: null,
   };
+
+  if (enrollments.errors) {
+    return {
+      ...defaultResponse,
+      onboardingStatus: 'Error while fetching data',
+    };
+  }
 
   // get most recent paid active enrollment
   const paidEnrollments = enrollments.filter((enrollment) => enrollment.is_active && (enrollment.mode === 'verified' || enrollment.mode === 'professional'));
--- a/src/users/data/api.test.js
+++ b/src/users/data/api.test.js
@@ -42,6 +42,18 @@ describe('API', () => {
     mockAdapter.reset();
   });
 
+  const enrollmentErrors = {
+    errors: [
+      {
+        code: null,
+        dismissible: true,
+        text: 'An unexpected error occurred. Please try refreshing the page.',
+        type: 'danger',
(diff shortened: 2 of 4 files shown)


src/users/data/api.js
```diff
@@ -33,10 +33,24 @@ export async function getEntitlements(username, page = 1) {
...
+      AppUrls.getEnrollmentsUrl(username),
+    );
+    return data;
+  } catch (error) {
+    return {
+      errors: [
+        {
+          code: null,
+          dismissible: true,
+          text: 'An unexpected error occurred. Please try refreshing the page.',
```
DawoudSheraz: maybe use error's own text instead of hardcoding the error text?

src/users/data/api.test.js
```diff
@@ -83,6 +95,13 @@ describe('API', () => {
       const response = await api.getOnboardingStatus(data, testUsername);
       expect(response).toEqual({ ...expectedSuccessResponse, onboardingStatus: 'Error while fetching data' });
     });
+
+    it('Unexpected error fetching enrollments', async () => {
+      // mockAdapter.onGet(url).reply(() => throwError(500, ''));
```
DawoudSheraz: what's going on here?

Ali-D-Akbar: Removing this.

src/users/data/api.js
Ali-D-Akbar: Okay.

src/users/enrollments/Enrollments.jsx
```diff
@@ -57,7 +68,7 @@ export default function Enrollments({
   });
 
   const tableData = useMemo(() => {
-    if (enrollmentData === null || enrollmentData.length === 0) {
+    if (enrollmentData === null || enrollmentData.errors || enrollmentData.length === 0) {
```
DawoudSheraz: where is enrollmentData.errors coming from?

DawoudSheraz (commented):
added some comments

src/users/enrollments/Enrollments.jsx
Ali-D-Akbar: from api when there's an error.

```
  try {
      AppUrls.getEnrollmentsUrl(username),
    );
    return data;
  } catch (error) {
    return {
      errors: [
        {
          code: null,
          dismissible: true,
          text: JSON.parse(error.customAttributes.httpErrorResponseData),
          type: 'danger',
          topic: 'enrollments',
        },
      ],
    };
  }
```

DawoudSheraz: but the error response is added to alert and not to enrollmentData.

src/users/data/api.js
```diff
@@ -33,10 +33,24 @@ export async function getEntitlements(username, page = 1) {
...
+      AppUrls.getEnrollmentsUrl(username),
+    );
+    return data;
+  } catch (error) {
+    return {
+      errors: [
+        {
+          code: null,
+          dismissible: true,
+          text: JSON.parse(error.customAttributes.httpErrorResponseData),
```
azanbinzahid, line 47: will it always have `customAttributes`? what-if it does not?

src/users/enrollments/Enrollments.jsx
Ali-D-Akbar: Oh, it is not needed here. I'll remove this.

src/users/data/api.js
Ali-D-Akbar: I'm not so sure about this but as far as I know, the standard error object in react has `customAttributes` extracted from the backend. So we may always have a `customAttributes` in `error`

DawoudSheraz: did you verify this on local?

Ali-D-Akbar: Yes.

DawoudSheraz (approved):

azanbinzahid (approved):

PR #144, opened 2021-08-30
refactor: self contain user summary and move modals

Ali-D-Akbar:
[PROD-2393](https://openedx.atlassian.net/browse/PROD-2393)
Moving all the modals out of the user summary to their separate components. No visual change.


DawoudSheraz (changes requested):
- Move all the actions in a folder called account-actions inside user
- add tests for each component. The tests are already in UserSummary. They just need to be moved appropriately.

src/users/account-actions/PasswordHistory.test.jsx
```diff
@@ -0,0 +1,46 @@
+import { mount } from 'enzyme';
+import React from 'react';
+
+import PasswordHistory from './PasswordHistory';
+import UserMessagesProvider from '../../userMessages/UserMessagesProvider';
+import UserSummaryData from '../data/test/userSummary';
+
+const PasswordHistoryWrapper = (props) => (
+  <UserMessagesProvider>
```
DawoudSheraz: we would not need message provider here. We are not using any alert component.

src/users/account-actions/ResetPassword.test.jsx
```diff
@@ -0,0 +1,58 @@
+import { mount } from 'enzyme';
+import React from 'react';
+
+import * as api from '../data/api';
+import ResetPassword from './ResetPassword';
+import UserMessagesProvider from '../../userMessages/UserMessagesProvider';
+import UserSummaryData from '../data/test/userSummary';
+
+const ResetPasswordWrapper = (props) => (
+  <UserMessagesProvider>
```
DawoudSheraz: same, no need for message provider.

src/users/account-actions/TogglePasswordStatus.test.jsx
```diff
@@ -0,0 +1,101 @@
+import { mount } from 'enzyme';
+import React from 'react';
+
+import * as api from '../data/api';
+import TogglePasswordStatus from './TogglePasswordStatus';
+import UserMessagesProvider from '../../userMessages/UserMessagesProvider';
+import UserSummaryData from '../data/test/userSummary';
+
+const TogglePasswordStatusWrapper = (props) => (
+  <UserMessagesProvider>
```
DawoudSheraz: and here too

DawoudSheraz (approved):
Address the nits in tests before merging.

src/users/account-actions/PasswordHistory.test.jsx
Ali-D-Akbar: Ah, I missed those. Removing them.

azanbinzahid (approved):
Toggling password status refreshes whole user data, is it expected? If yes then it is good to go.

Ali-D-Akbar:
> Toggling password status refreshes whole user data, is it expected? If yes then it is good to go.

Yes, it is the expected. We will revisit this probably in the near future.

PR #145, opened 2021-09-05
feat: add sso record component v2

azanbinzahid:
## Description
This PR adds Card and List component for SSO Records as part of Support Tool Redesign.

- Uses new Table variant for rendering data Table
- width set to 100% per card as per design
- table width set to auto to avoid table consuming the entire card
- Single card per row

## Linked Ticket
[PROD-2459](https://openedx.atlassian.net/browse/PROD-2459)
[PROD-2460](https://openedx.atlassian.net/browse/PROD-2460)

## Screenshots
Screenshot 2021-09-14 at 9 48 45 PM

Screenshot 2021-09-14 at 10 16 31 PM

## How to test locally:
1. Add List Component 
```js
      <SingleSignOnRecordsList
        username={data.user.username}
      />
```
2. Add dummy data here: `/admin/social_django/usersocialauth/`
3. Populate extra data (you may add response from test data file) 

## TODO
- [x] add instructions to test components locally 
- [x] get initial review
- [x] write component test

### DIFF
--- a/src/users/data/api.js
+++ b/src/users/data/api.js
@@ -60,10 +60,20 @@ export async function getSsoRecords(username) {
     );
     let parsedData = [];
     if (data.length > 0) {
-      parsedData = data.map((entry) => ({
-        ...entry,
-        extraData: JSON.parse(entry.extraData),
-      }));
+      parsedData = data.map((entry) => {
+        const extraData = JSON.parse(entry.extraData);
+
+        Object.keys(extraData).forEach((key) => {
+          extraData[key] = Array.isArray(extraData[key]) && extraData[key].length > 0
+            ? extraData[key][0]
+            : extraData[key];
+        });
+
+        return {
+          ...entry,
+          extraData,
+        };
+      });
     }
     return parsedData;
   } catch (error) {
--- a/src/users/v2/CopyShowHyperLinks.jsx
+++ b/src/users/v2/CopyShowHyperLinks.jsx
@@ -1,0 +1,40 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import { Hyperlink } from '@edx/paragon';
+
+export default function CopyShowHyperlinks({ text }) {
+  const [copyText, setCopyText] = useState('Copy');
+
+  return (
+    <>
+      <Hyperlink
+        destination=""
+        target="_blank"
+        onClick={(e) => {
+          e.preventDefault();
+          navigator.clipboard.writeText(text);
+          // show temp check mark after copy
+          setCopyText('Copy\u2713');
+          setInterval(() => setCopyText('Copy'), 3000);
+        }}
+      >
+        {copyText}
+      </Hyperlink>
+      <Hyperlink
+        destination=""
+        target="_blank"
+        onClick={(e) => {
+          e.preventDefault();
+          // eslint-disable-next-line no-alert
+          alert(text);
+        }}
+      >
+        Show
+      </Hyperlink>
+    </>
+  );
+}
+
+CopyShowHyperlinks.propTypes = {
+  text: PropTypes.string.isRequired,
+};
--- a/src/users/v2/SingleSignOnRecordCard.jsx
+++ b/src/users/v2/SingleSignOnRecordCard.jsx
@@ -1,0 +1,81 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {
+  Card, Row, Col,
+} from '@edx/paragon';
+import TableV2 from '../../components/Table';
+import { formatDate } from '../../utils';
+import CopyShowHyperlinks from './CopyShowHyperLinks';
+
+export default function SingleSignOnRecordCard({ ssoRecord }) {
+  let data;
+  let columns;
+
+  if (ssoRecord != null) {
+    data = { ...ssoRecord.extraData };
+    columns = React.useMemo(
+      () => Object.keys(data)
+        .sort((a, b) => a > b)
+        .map((key) => ({
+          Header: key,
+          accessor: key,
+        })),
+    );
+
+    Object.keys(data).forEach((key) => {
+      const value = data[key].toString();
+      if (value.length > 14) {
+        data[key] = <CopyShowHyperlinks text={value} />;
+      }
+    });
+    data = React.useMemo(
+      () => data,
+    );
+  }
+
+  return (
+    ssoRecord ? (
+      <Card className="pt-2 px-3 mb-1 w-100">
+        <Card.Body className="p-0">
+          <Card.Title as="h3" className="btn-header mt-4">
+            {ssoRecord.provider}
+          </Card.Title>
+          <Row>
+            <Col>
+              <Card.Subtitle align="left" as="h4">
+                {ssoRecord.uid}
+              </Card.Subtitle>
+            </Col>
+            <Col>
+              <Card.Subtitle align="right" as="h4">
+                Last Modified: {formatDate(ssoRecord.modified)}
+              </Card.Subtitle>
+            </Col>
+          </Row>
+
+          <Card.Title as="h5" className="btn-header mt-4">
+            Additional Data
+          </Card.Title>
+
+          <TableV2
+            styleName="sso-table"
+            id="sso-data-new"
+            data={[data]}
+            columns={columns}
+          />
+        </Card.Body>
+      </Card>
+    ) : (
+      <></>
+    )
+  );
+}
+
+SingleSignOnRecordCard.propTypes = {
+  ssoRecord: PropTypes.shape({
+    provider: PropTypes.string,
+    uid: PropTypes.string,
+    modified: PropTypes.string,
+    extraData: PropTypes.object,
+  }).isRequired,
+};
--- a/src/users/v2/SingleSignOnRecordCard.test.jsx
+++ b/src/users/v2/SingleSignOnRecordCard.test.jsx
@@ -1,0 +1,72 @@
+import { mount } from 'enzyme';
+import React from 'react';
+import { camelCaseObject } from '@edx/frontend-platform';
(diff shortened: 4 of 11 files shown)


src/users/v2/SingleSignOnRecordCard.jsx
```diff
@@ -0,0 +1,116 @@
...
+
+function createHyperlinks(value) {
+  return (
+    <>
+      <Hyperlink
+        destination=""
+        target="_blank"
+        onClick={(e) => {
+          e.preventDefault();
+          navigator.clipboard.writeText(value);
```
DawoudSheraz: does this give a visual indication that copy action has taken place?

src/users/v2/SingleSignOnRecordCard.jsx
```diff
@@ -0,0 +1,116 @@
...
+        }}
+      >
+        Copy
+      </Hyperlink>
+      <Hyperlink
+        destination=""
+        target="_blank"
+        onClick={(e) => {
+          e.preventDefault();
+          alert(value); // Modal can be used instead of alert()
```
DawoudSheraz: let's keep this alert for now.

src/users/v2/SingleSignOnRecordCard.jsx
```diff
@@ -0,0 +1,116 @@
...
+            </Row>
+
+            <Card.Title as="h5" className="btn-header mt-4">
+              Additional Data
+            </Card.Title>
+
+            <Table
+              className="table-responsive h5"
+              id="sso-data"
+              data={[data]}
```
DawoudSheraz: Why is data conversion to a list of dict necessary?

src/users/v2/SingleSignOnRecordCard.jsx
azanbinzahid: just added a checkmark for few seconds after text has been copied.

src/users/v2/SingleSignOnRecordCard.jsx
azanbinzahid: Table accepts data as list of dict.

src/users/v2/CopyShowHyperLinks.jsx
```diff
@@ -0,0 +1,39 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import { Hyperlink } from '@edx/paragon';
+
+export default function CopyShowHyperlinks({ textValue }) {
```
DawoudSheraz: nit: either text or value would be good.

DawoudSheraz (commented):
codewise, the changes look good. I will test it out on local as well. You can work on adding tests.

src/users/v2/SingleSignOnRecordsList.jsx
```diff
@@ -0,0 +1,47 @@
...
+import { camelCaseObject } from '@edx/frontend-platform';
+import PropTypes from 'prop-types';
+import { Row, Col } from '@edx/paragon';
+import UserMessagesContext from '../../userMessages/UserMessagesContext';
+import { getSsoRecords } from '../data/api';
+import PageLoading from '../../components/common/PageLoading';
+import AlertList from '../../userMessages/AlertList';
+import SingleSignOnRecordCard from './SingleSignOnRecordCard';
+
+export default function SingleSignOnRecords({ username }) {
```
Ali-D-Akbar: Just for the consistency, can we name the file `SingleSignOnRecords.jsx` instead?

Ali-D-Akbar (commented):
The changes look good to me as well. I just need to verify the UI on local. 👍

src/users/data/api.js
```diff
@@ -60,10 +60,20 @@ export async function getSsoRecords(username) {
...
+          extraData[key] = Array.isArray(extraData[key]) && extraData[key].length > 0
+            ? extraData[key][0]
+            : extraData[key];
+        });
+
+        return {
+          ...entry,
+          extraData,
+        };
+      });
```
DawoudSheraz, line 76: why the return statement here? parsedData is by default []. How will that affect the parsing and existing SSO component?

src/users/v2/SingleSignOnRecordCard.test.jsx
```diff
@@ -0,0 +1,74 @@
...
+import { formatDate } from '../../utils';
+
+describe('Onboarding Status', () => {
+  // prepare data
+  const ssoRecords = camelCaseObject(ssoRecordsData.map((entry) => ({
+    ...entry,
+    extraData: JSON.parse(entry.extraData),
+  })));
+
+  for (let index = 0; index < ssoRecords.length; index++) {
```
DawoudSheraz: why use for loop? You can use jest test.each to repeat the tests by modifying the data. Ref: https://jestjs.io/docs/api#testeachtablename-fn-timeout

Also, using for loop to wrap a lot of test operations seems tricky. Do you have any examples/resources that advocate for this approach?

src/users/data/api.js
```diff
@@ -60,10 +60,20 @@ export async function getSsoRecords(username) {
...
-      }));
+      parsedData = data.map((entry) => {
+        const extraData = JSON.parse(entry.extraData);
+
+        Object.keys(extraData).forEach((key) => {
+          extraData[key] = Array.isArray(extraData[key]) && extraData[key].length > 0
+            ? extraData[key][0]
+            : extraData[key];
+        });
+
```
DawoudSheraz, line 71: won't we need to update the existing SSO component to accommodate this change??

src/users/v2/SingleSignOnRecordCard.jsx
```diff
@@ -0,0 +1,81 @@
...
+          value,
+        };
+      } else {
+        data[key] = value;
+      }
+    });
+  }
+
+  return (
+    ssoRecord ? (
```
Ali-D-Akbar, line 37: Are we not going to show the loader here? This approach won't even show the column headings incase of no ssoRecords.

src/users/v2/SingleSignOnRecordsList.jsx
```diff
@@ -0,0 +1,47 @@
...
+import { camelCaseObject } from '@edx/frontend-platform';
+import PropTypes from 'prop-types';
+import { Row, Col } from '@edx/paragon';
+import UserMessagesContext from '../../userMessages/UserMessagesContext';
+import { getSsoRecords } from '../data/api';
+import PageLoading from '../../components/common/PageLoading';
+import AlertList from '../../userMessages/AlertList';
+import SingleSignOnRecordCard from './SingleSignOnRecordCard';
+
+export default function SingleSignOnRecordsList({ username }) {
```
Ali-D-Akbar: As said before, it'd be better to keep `SingleSignOnRecords.jsx` for the consistency: https://github.com/edx/frontend-app-support-tools/pull/145#discussion_r704145615

Ali-D-Akbar (commented):
Please rebase this branch onto master and add the component in UserSummary.jsx. It is kind of hard to verify the changes and the use-cases on local without it.

src/users/v2/SingleSignOnRecordCard.jsx
DawoudSheraz: Column headings are not fixed. They depend upon the received data from the API.

src/users/v2/SingleSignOnRecordsList.jsx
ansabgillani: React filename/component naming convention not followed:
Component should be renamed to SingleSignOnRecordsList

Ali-D-Akbar (changes requested):
1. There seem to have some overlapping text in this:
Screenshot 2021-09-13 at 8 09 25 PM

2. Two SSO cards in one row? Isn't one SSO record each row the right way according to AC?
Screenshot 2021-09-13 at 8 09 37 PM

3. Too much spacing above and below the fields in the 2nd screenshot.
4. In the AC, we see a card with only lower border. It'd be better to keep the consistency here as both enrollment and entitlement components use the same approach
image
5. Text size also seems to be larger. Use `size="sm"` variant or `font-size: small;` style in the parent component.

DawoudSheraz:
@Ali-D-Akbar 
1. Yes, no overlapping text.
2. what width settings were used? Ideally, one record/card per row.
3. Can't quite understand this point.
4. Enrollments and Entitlements are using Table. SSO can't use table because extra data/metadata varies from provider. SSO is using card component, so that is why there will be extra border. 
5. Noted.

ansabgillani:
@DawoudSheraz Clarification on 3. 
Consider the top and bottom padding on the either. 

How it is supposed to look: 
Screenshot at Sep 13 22-07-52

How it actually looks:
Screenshot at Sep 13 22-05-04

ansabgillani:
Can you please verify if following is okay as it is or not:

1. Left align mismatch of the "Additional Data" and the elements in the table.
2. Proper alignment of copy and show to left and right ends respectively of their columns.
3. Small nits like bold texts and text fonts are yet to be addressed.
4. Note that in previous screenshot, "PM" is going on the next line instead of same line despite the same number of characters causing an extra bit of margin space.

src/users/v2/SingleSignOnRecordCard.test.jsx
```diff
@@ -0,0 +1,72 @@
+import { mount } from 'enzyme';
+import React from 'react';
+import { camelCaseObject } from '@edx/frontend-platform';
+import { waitForComponentToPaint } from '../../setupTest';
+import ssoRecordsData from '../data/test/ssoRecords';
+import SingleSignOnRecordCard from './SingleSignOnRecordCard';
+import { formatDate } from '../../utils';
+
+describe.each(ssoRecordsData)('SSO Records', (ssoRecordData) => {
```
Ali-D-Akbar: nit: Single Sign on Record Card.

Ali-D-Akbar (approved):
Looks good to me. 👍

PR #146, opened 2021-09-09
feat: add entitlementV2 component

Ali-D-Akbar:
[PROD-2463](https://openedx.atlassian.net/browse/PROD-2463)

This PR adds a new design for Entitlement Component while keeping the older components. This v2 component uses the same Table component that we used in Enrollment v2 component.

Entitlement with Expanded Row:
Screenshot 2021-09-09 at 11 32 36 PM

Entitlement with Disabled Reissue and Enabled Expire actions:
Screenshot 2021-09-09 at 11 33 06 PM

### DIFF
--- a/src/users/entitlements/v2/Entitlements.jsx
+++ b/src/users/entitlements/v2/Entitlements.jsx
@@ -1,0 +1,302 @@
+import React, {
+  useMemo, useState, useCallback, useRef, useLayoutEffect, useContext, useEffect,
+} from 'react';
+import PropTypes from 'prop-types';
+
+import {
+  Button, TransitionReplace, Dropdown, Hyperlink,
+} from '@edx/paragon';
+import { camelCaseObject, getConfig } from '@edx/frontend-platform';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faMinus, faPlus } from '@fortawesome/free-solid-svg-icons';
+import EntitlementForm from '../EntitlementForm';
+import { CREATE, REISSUE, EXPIRE } from '../EntitlementActions';
+import PageLoading from '../../../components/common/PageLoading';
+import CourseSummary from '../../courseSummary/CourseSummary';
+import TableV2 from '../../../components/Table';
+import { getEntitlements } from '../../data/api';
+import UserMessagesContext from '../../../userMessages/UserMessagesContext';
+import { formatDate } from '../../../utils';
+import AlertList from '../../../userMessages/AlertList';
+
+export default function Entitlements({
+  changeHandler, user,
+}) {
+  const { add, clear } = useContext(UserMessagesContext);
+  const [formType, setFormType] = useState(null);
+  const [userEntitlement, setUserEntitlement] = useState(undefined);
+  const [courseSummaryUUID, setCourseSummaryUUID] = useState(null);
+  const [entitlementData, setEntitlementData] = useState(null);
+  const formRef = useRef(null);
+  const summaryRef = useRef(null);
+
+  useEffect(() => {
+    clear('entitlements');
+    getEntitlements(user).then((result) => {
+      const camelCaseResult = camelCaseObject(result);
+      if (camelCaseResult.errors) {
+        camelCaseResult.errors.forEach(error => add(error));
+        setEntitlementData({ results: [] });
+      } else {
+        setEntitlementData(camelCaseResult);
+      }
+    });
+  }, [user]);
+
+  const tableData = useMemo(() => {
+    if (entitlementData === null) {
+      return [];
+    }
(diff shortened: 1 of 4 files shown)


src/users/entitlements/v2/Entitlements.jsx
```diff
@@ -0,0 +1,298 @@
...
+          href={`${getConfig().LMS_BASE_URL}/courses/${entitlement.enrollmentCourseRun}`}
+          rel="noopener noreferrer"
+          target="_blank"
+        >
+          {entitlement.enrollmentCourseRun}
+        </a>
+      ) : 'Course Run Not Selected'),
+      expiredAt: formatDate(entitlement.expiredAt),
+      created: formatDate(entitlement.created),
+      modified: formatDate(entitlement.modified),
```
DawoudSheraz: keep the sequence and column titles the same as the design.

src/users/entitlements/v2/Entitlements.jsx
```diff
@@ -0,0 +1,298 @@
...
+            Actions
+          </Dropdown.Toggle>
+          <Dropdown.Menu>
+            <Dropdown.Item
+              onClick={() => {
+                setCourseSummaryUUID(null);
+                setUserEntitlement(entitlement);
+                setFormType(REISSUE);
+              }}
+              disabled={Boolean(!entitlement.enrollmentCourseRun)}
```
DawoudSheraz, line 107: Does disable work with item? Instead of disabling it, we can skip the item altogether if the action is not applicable.

src/users/entitlements/v2/Entitlements.jsx
```diff
@@ -0,0 +1,298 @@
...
+              className="small"
+            >
+              Reissue
```
DawoudSheraz, line 110: Lets use full words, like Reissue Entitlement

DawoudSheraz (commented):
the changes look good on the surface. I have not tested it on local yet. Address the suggested changes before the final approval.

src/users/entitlements/v2/Entitlements.jsx
Ali-D-Akbar: Hmm, I think it's pretty obvious that this action is specific to an entitlement.

src/users/entitlements/v2/Entitlements.jsx
Ali-D-Akbar: This is how it works. Reissue is disabled and expired is enabled. I think it still makes sense to keep the action there just so the support user knows how many actions are possible on each entitlement.

Screenshot 2021-09-09 at 11 25 44 PM

src/users/entitlements/v2/Entitlements.jsx
DawoudSheraz: Yes, it is. But take an example of the course dashboard. We see the Resume Course button there. Only Resume would have been enough too. Having explicit words is better.

src/index.jsx
```diff
@@ -36,7 +36,7 @@ subscribe(APP_READY, () => {
         <Switch>
           <Route exact path="/" component={SupportHomePage} />
           <Route path="/users" component={UserPage} />
-          {/* <Route path="/usersv2" component={UserPageV2} /> */}
+          <Route path="/usersv2" component={UserPageV2} />
```
DawoudSheraz: comment this out.

DawoudSheraz (changes requested):
Enrollment Date and Course Run ID are two distinct headers. The course run ID column is missing.

Ali-D-Akbar:
> Enrollment Date and Course Run ID are two distinct headers. The course run ID column is missing.

Screenshot 2021-09-10 at 11 21 10 PM
Added course run ID and removed enrollment date for now as we don't receive this from backend.

src/users/entitlements/v2/Entitlements.jsx
```diff
@@ -0,0 +1,302 @@
...
+        Header: 'Unenrolled Run', accessor: 'unenrolledRun', sortable: true,
+      },
+    ],
+    [],
+  );
+
+  const renderRowSubComponent = useCallback(
+    ({ row }) => (
+      <TableV2
+        // eslint-disable-next-line react/prop-types
```
DawoudSheraz, line 227: is this necessary?

DawoudSheraz (approved):

DawoudSheraz:
squash the commits before merging.

src/users/entitlements/v2/Entitlements.jsx
Ali-D-Akbar: Yes, react/prop-types eslint rule says every function must have their prop-types defined. This is a callback so we may not know what each row has.

PR #149, opened 2021-09-21
feat: add feature based enrollment components

DawoudSheraz:
### Description
Adds the component to get feature-based enrollment details for a particular course. To test out the component,

1. Uncomment FBE route and component import from index.js
2. Create a course in studio. Make sure the course has both audit and verified modes(http://localhost:18000/admin/course_modes/coursemode/)
3. Set the course start date in past
4. Navigate to http://localhost:18000/admin/content_type_gating/contenttypegatingconfig/add/.
Create a record at the site level with Enabled = Yes and Enabled As Of set to some date in the past.
5. Repeat the last step for http://localhost:18000/admin/course_duration_limits/coursedurationlimitconfig/add/

First verify that the FBE api(http://localhost:18000/support/feature_based_enrollment_details/<course_id>) returns some response instead of being an empty dict. If the empty dict is returned, try associating the created course with the configs added in 4 & 5. There is a field present in that config model. 

**NOTE: There is a small issue with FBE component at the moment. When the course id is added in the search bar, search button is clicked and a valid response is returned, clicking the search button again without changing the value does not result in any api call. The child component triggers useEffect only when the course id has changed. This will be fixed in a followup PR**

### Things to verify

- Valid/Invalid course ids behavior
- addition/update of query param in the url when changing the value
- adding a query param manually, refreshing the page, and make sure input field has that value

### Screenshots
Screenshot 2021-09-22 at 10 21 04 PM

Screenshot 2021-09-22 at 10 21 29 PM

Screenshot 2021-09-22 at 10 22 21 PM

### DIFF
--- a/src/FeatureBasedEnrollments/FeatureBasedEnrollment.jsx
+++ b/src/FeatureBasedEnrollments/FeatureBasedEnrollment.jsx
@@ -1,0 +1,67 @@
+import React, {
+  useEffect,
+  useContext,
+  useState,
+} from 'react';
+import { camelCaseObject } from '@edx/frontend-platform';
+import { Row, Col } from '@edx/paragon';
+import PropTypes from 'prop-types';
+
+import UserMessagesContext from '../userMessages/UserMessagesContext';
+import AlertList from '../userMessages/AlertList';
+import getFeatureBasedEnrollmentDetails from './data/api';
+import PageLoading from '../components/common/PageLoading';
+import FeatureBasedEnrollmentCard from './FeatureBasedEnrollmentCard';
+
+export default function FeatureBasedEnrollment({ courseId }) {
+  const [fbeData, setFBEData] = useState(undefined);
+  const { add, clear } = useContext(UserMessagesContext);
+
+  useEffect(() => {
+    clear('featureBasedEnrollment');
+    setFBEData(undefined);
+    getFeatureBasedEnrollmentDetails(courseId).then((response) => {
+      const camelCaseResponse = camelCaseObject(response);
+      if (camelCaseResponse.errors) {
+        camelCaseResponse.errors.forEach(error => add(error));
+      } else {
+        setFBEData(camelCaseResponse);
+      }
+    });
+  }, [courseId]);
+
+  return (
+    <section className="mb-3">
+      <AlertList topic="featureBasedEnrollment" className="mb-3" />
+      <h3 id="fbe-title-header" className="my-4">Feature Based Enrollment Configuration</h3>
+
+      {/* eslint-disable-next-line no-nested-ternary */}
+      {fbeData
+        ? (
+          (Object.keys(fbeData).length > 0
+            ? (
+              <>
+                <h4 className="my-4">Course Title: {fbeData.courseName}</h4>
+                <Row>
+                  <Col>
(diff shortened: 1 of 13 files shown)


azanbinzahid (approved):
Tested locally, working as expected. A small concern over the table in card, we can make headings bold or put colons at the end of them.

src/FeatureBasedEnrollments/FeatureBasedEnrollment.jsx
```diff
@@ -0,0 +1,68 @@
...
+  const [fbeData, setFBEData] = useState(undefined);
+  const { add, clear } = useContext(UserMessagesContext);
+
+  useEffect(() => {
+    clear('featureBasedEnrollment');
+    setFBEData(undefined);
+    getFeatureBasedEnrollmentDetails(courseId).then((response) => {
+      const camelCaseResponse = camelCaseObject(response);
+      if (camelCaseResponse.errors) {
+        clear('featureBasedEnrollment');
```
Ali-D-Akbar: What's the reason to make another clear call here?

src/FeatureBasedEnrollments/FeatureBasedEnrollmentIndexPage.jsx
```diff
@@ -0,0 +1,113 @@
...
+  const handleSearchInput = useCallback((inputValue) => {
+    if (inputValue !== undefined && inputValue !== '') {
+      clear('featureBasedEnrollment');
+      if (!validateInput(inputValue)) {
+        return;
+      }
+      setSearchValue(inputValue);
+      pushHistoryIfChanged(`/feature_based_enrollments/?course_id=${inputValue}`);
+    } else if (inputValue === '') {
+      clear('featureBasedEnrollment');
```
Ali-D-Akbar: can be recuded to one clear call as well.

src/FeatureBasedEnrollments/FeatureBasedEnrollmentIndexPage.jsx
```diff
@@ -0,0 +1,113 @@
...
+  }, []);
+
+  return (
+    <main className="ml-5 mr-5 mt-3 mb-5">
+
+      <section className="mb-3">
+        <Link to="/">&lt; Back to Tools</Link>
+      </section>
+
+      <AlertList topic="featureBasedEnrollment" className="mb-3" />
```
Ali-D-Akbar: Wouldn't that make duplicate alerts? There's one in the `FeatureBasedEnrollment.jsx` as well.

Ali-D-Akbar (commented):
Verifying them on local.

src/FeatureBasedEnrollments/FeatureBasedEnrollment.jsx
DawoudSheraz: yeah, that is redundant.

src/FeatureBasedEnrollments/FeatureBasedEnrollmentIndexPage.jsx
DawoudSheraz: Yes, good catch.

DawoudSheraz:
> Tested locally, working as expected. A small concern over the table in card, we can make headings bold or put colons at the end of them.

Fair point, I will update the styles.

Ali-D-Akbar (approved):
👍

PR #163, opened 2021-10-07
feat: add collapse all button

Ali-D-Akbar:
[PROD-2501](https://openedx.atlassian.net/browse/PROD-2501)

Replaces `Expand All` with `Collapse All` button When all the rows are in expanded state.

### DIFF
--- a/src/users/enrollments/v2/Enrollments.jsx
+++ b/src/users/enrollments/v2/Enrollments.jsx
@@ -106,13 +106,14 @@ export default function Enrollments({
     }
   });
 
-  const expandAllRowsHandler = ({ getToggleAllRowsExpandedProps }) => (
+  const expandAllRowsHandler = ({ getToggleAllRowsExpandedProps, isAllRowsExpanded }) => (
     <a {...getToggleAllRowsExpandedProps()} className="link-primary">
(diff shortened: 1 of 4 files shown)


src/users/enrollments/v2/Enrollments.jsx
```diff
@@ -106,13 +106,14 @@ export default function Enrollments({
     }
   });
 
-  const expandAllRowsHandler = ({ getToggleAllRowsExpandedProps }) => (
+  const expandAllRowsHandler = ({ getToggleAllRowsExpandedProps, isAllRowsExpanded }) => (
```
DawoudSheraz, line 109: is `isAllRowsExpanded` defined by react-table? If not, can we rename it to areAllRowsExpanded.

DawoudSheraz (approved):

src/users/enrollments/v2/Enrollments.jsx
Ali-D-Akbar: It is defined in react-table. Cannot rename it. https://react-table.tanstack.com/docs/api/useExpanded#instance-properties

azanbinzahid (approved):

PR #164, opened 2021-10-11
fix: reissue entitlement on course run

Ali-D-Akbar:
[PROD-2536](https://openedx.atlassian.net/browse/PROD-2536)
Revisiting entitlement reissue. Allowing Reissue only on an entitlement that is not expired and has a course run associated with it.

### DIFF
--- a/src/users/entitlements/v2/Entitlements.jsx
+++ b/src/users/entitlements/v2/Entitlements.jsx
@@ -107,7 +107,8 @@ export default function Entitlements({
                 setUserEntitlement(entitlement);
                 setFormType(REISSUE);
               }}
-              disabled={Boolean(entitlement.enrollmentCourseRun || !entitlement.expiredAt)}
+              // disable if entitlement has no associated course run or the entitlement is expired
+              disabled={Boolean(!entitlement.enrollmentCourseRun || entitlement.expiredAt)}
               className="small"
             >
               Reissue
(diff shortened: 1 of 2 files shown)


src/users/entitlements/v2/Entitlements.test.jsx
```diff
@@ -135,36 +165,6 @@ describe('Entitlements V2 Listing', () => {
     });
   });
 
-  describe('Expire Entitlement button', () => {
```
Ali-D-Akbar, line 138: Expire tests were failing for some reason. Changing the order of tests here fixed the problem for now but will take a look at it afterwards on why they were failing.

src/users/entitlements/v2/Entitlements.jsx
```diff
@@ -107,7 +107,7 @@ export default function Entitlements({
                 setUserEntitlement(entitlement);
                 setFormType(REISSUE);
               }}
-              disabled={Boolean(entitlement.enrollmentCourseRun || !entitlement.expiredAt)}
+              disabled={Boolean(!entitlement.enrollmentCourseRun || entitlement.expiredAt)}
```
azanbinzahid, line 111: since its a trivial check, can we add a comment for future ref?

src/users/entitlements/v2/Entitlements.test.jsx
azanbinzahid: It may lead to failing tests arbitrarily, creating uncertainty. Can we look into this in same PR?

Ali-D-Akbar: It has to do with the data being mocked in an order. Mocking reissue before expired changed the results for some reason. (seems like the wrapper isn't unmount after each test is run, has to do something with that extra describe inside describe) That being said, the tests aren't flaky anymore.

azanbinzahid (approved):

PR #165, opened 2021-10-15
fix: FBE same course consecutive search

ansabgillani:
### Description

- Fixing API call on search click with the same course

### DIFF
--- a/src/FeatureBasedEnrollments/FeatureBasedEnrollmentIndexPage.test.jsx
+++ b/src/FeatureBasedEnrollments/FeatureBasedEnrollmentIndexPage.test.jsx
@@ -18,13 +18,18 @@ const FeatureBasedEnrollmentIndexPageWrapper = (props) => (
 );
 
 describe('Feature Based Enrollment Index Page', () => {
-  let location; let wrapper;
+  let location; let wrapper; let apiMock;
   const courseId = 'course-v1:testX+test123+2030';
 
   beforeEach(() => {
     location = { pathname: '/v2/feature_based_enrollments', search: '' };
   });
 
+  afterEach(() => {
+    if (apiMock) {
+      apiMock.mockReset();
+    }
+  });
   it('default page render', async () => {
     wrapper = mount(<FeatureBasedEnrollmentIndexPageWrapper location={location} />);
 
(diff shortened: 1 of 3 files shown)


Ali-D-Akbar (approved):
Fair solution for the problem but can we do something other than sending a prop in FeatureBasedEnrollment component? It seems to add an extra unnecessary layer of dependability whenever we use it.

ansabgillani:
> Fair solution for the problem but can we do something other than sending a prop in FeatureBasedEnrollment component? It seems to add an extra unnecessary layer of dependability whenever we use it.

The idea behind this solution was to get a signal from the FBEIndexPage (Parent) to FBE (Child) component so it can make the API call whenever the search button was clicked. I don't see any other workaround than props to pass the signal.

azanbinzahid:
Just a thought, what if we re-render the component from [index page](https://github.com/edx/frontend-app-support-tools/blob/master/src/FeatureBasedEnrollments/FeatureBasedEnrollmentIndexPage.jsx) based on loading state instead of passing it as a prop?

ansabgillani:
> Just a thought, what if we re-render the component from [index page](https://github.com/edx/frontend-app-support-tools/blob/master/src/FeatureBasedEnrollments/FeatureBasedEnrollmentIndexPage.jsx) based on loading state instead of passing it as a prop?

@azanbinzahid tried this but did not work. The problem I am facing is that the API call is being called in the child component and should be dependent on a signal from the parent. Even if I re-render the component, the useEffect still requires a signal from the parent.

DawoudSheraz (changes requested):
- Missing related unit test
- It seems the loading state is not loading but acting as a toggle so that useEffect can be triggered. We should rename it so that it does not give an impression of loading data.

ansabgillani:
Follow up on this. According to the react way, the best and most convenient way to approach this problem is to form up a signal that fires the useEffect triggering an API call in the child component. 

https://github.com/facebook/react/issues/6646#issuecomment-215580357

src/FeatureBasedEnrollments/FeatureBasedEnrollmentIndexPage.test.jsx
```diff
@@ -63,12 +67,32 @@ describe('Feature Based Enrollment Index Page', () => {
...
     expect(history.push).toHaveBeenCalledWith(`/v2/feature_based_enrollments/?course_id=${courseId}`);
 
-    apiMock.mockReset();
+    history.push.mockReset();
+  });
+
+  it('api call made on each click', async () => {
+    apiMock = jest
+      .spyOn(api, 'default')
+      .mockImplementation(() => Promise.resolve(fbeEnabledResponse));
```
DawoudSheraz: nit: keep this all on the same line and only extend if the line length is violated.

DawoudSheraz (approved):
- Squash the commits and rebase the branch
- Address one nit

azanbinzahid (approved):

PR #168, opened 2021-10-29
feat: using edx/reactifex for push translations

DawoudSheraz:
### Description

- Add edx/reactifex as dependency and use Transifex api v3 scripts for push translation
- Update frontend-platform to the latest version to be compatible with edx/reactifex
- Add 2 FormattedMessage elements for testing the push translation
- Add Transifex config file

Note that the push translations will be tested through the Jenkins job.

### Links

- https://github.com/edx/frontend-platform/blob/master/docs/how_tos/i18n.rst
- https://openedx.atlassian.net/wiki/spaces/LOC/pages/3103293538/MFE+Translation+How+To+s
- https://openedx.atlassian.net/wiki/spaces/IM/pages/3219324929/Using+edx+reactifex+in+push+translation+script+for+MFEs


package.json
```diff
@@ -45,6 +45,7 @@
     "prop-types": "15.7.2",
     "react": "16.14.0",
     "react-dom": "16.14.0",
+    "react-intl": "^5.21.0",
```
davidjoy: This is a dependency of frontend-platform, so it doesn't need to be added in the app itself.

davidjoy (changes requested):

package.json
DawoudSheraz: I was getting both unit tests failure and eslint unresolved import error if I was not adding it as a dependency. Some other repos([ora-grading](https://github.com/edx/frontend-app-ora-grading/blob/7e99e7cca29568afcb4347f587411aed6c1e9e7f/package.json), [special-exams](https://github.com/edx/frontend-lib-special-exams/blob/f0eaf4faad48831f17876d8bc47c14d46bdcc92d/package.json)) had both frontend-platform and react-intl as dependency.

davidjoy: I'm not sure what ora-grading and special-exams are doing, but they shouldn't have react-intl as a direct dependency.  That sounds like a mistake as well.  Everything should be imported from `@edx/frontend-platform/i18n`, rather than from `react-intl`.  frontend-platform re-exports everything because we have some shims and wrappers around some of the components, and also because you can only have one version of react-intl running at a time or it won't work.

Any imports from `'react-intl`' should be replaced with `'@edx/frontend-platform/i18n'`, even in tests, and then the dependency can be removed.

DawoudSheraz: Removed `react-intl` as a dependency and importing the provider from frontend-platform now.

src/users/account-actions/v2/ResetPassword.jsx
```diff
@@ -41,10 +42,16 @@ export default function ResetPassword({
...
-                Do you wish to proceed?
-              </p>
+              <FormattedMessage
+                id="supportTools.accountActions.resetPassword"
+                tagName="p"
+                description="Password reset email instructions prompt"
+                defaultMessage="We will send a message with password recovery instructions to the email address {email}.
+                Do you wish to proceed?"
+                values={{
+                  email: <b>{email}</b>,
```
azanbinzahid, line 52: will it pick b tag automatically?

azanbinzahid: Will it pick `b` tag automatically given we are providing `tagName` separately?

azanbinzahid (approved):

src/users/account-actions/v2/ResetPassword.jsx
DawoudSheraz: Yes, **values** contain the individual value for the variables defined in the message.

ansabgillani (approved):

Ali-D-Akbar (approved):
Are these all the messages in Support Tools that are subjected to translation?

DawoudSheraz:
> Are these all the messages in Support Tools that are subjected to translation?

There might be 1-2 more but they will be added/tested in a follow-up PR to verify the new scripts also work well with existing and new translation strings.

src/users/account-actions/v2/ResetPassword.jsx
davidjoy: Nit: I think we want to avoid using `<b>` and use `<strong>` instead - the latter carries semantic meaning for accessibility devices, which I think we want here.

DawoudSheraz:
@davidjoy can you take a look, please? It seems I can't merge the PR unless the requested changes are addressed, which I have done already.

davidjoy (approved):

PR #169, opened 2021-10-30
fix: sort by datevalue instead of formatted string

azanbinzahid:
## Description
In `TableV2`, the cell values are being formatted at data level whereas it has to be formatted only at display/rendering level. This is causing sort issues because sorting is currently being done on formatted string value but not actual value. This PR attempts to fix this on Enrollment Component 

## Linked Ticket
[PROD-2549](https://openedx.atlassian.net/browse/PROD-2549)

### Date: 
Before | After
:-------------------------:|:-------------------------:
image | image

### Hyperlink
Before | After
:-------------------------:|:-------------------------:
image | image

## References:
- prop validation issue [link](https://spectrum.chat/react-table/general/eslint-prop-types-problem-with-cell~b51c7ae7-271f-45c6-b7a0-0cdde4c89c74)
- column sorting works better if in raw format (i.e. not in display/render) [link](https://stackoverflow.com/questions/63927644/how-does-one-supply-a-custom-sort-function-for-react-table-7)
- Cell Implementation [link](https://stackoverflow.com/questions/57735811/manipulate-date-inside-react-table)

## TODO:
- [x] add comparison SS
- [ ] If this fix is accepted, it has to be repeated at every component using Table V2.

### DIFF
--- a/src/users/enrollments/v2/Enrollments.jsx
+++ b/src/users/enrollments/v2/Enrollments.jsx
@@ -58,16 +58,16 @@ export default function Enrollments({
     }
     return enrollmentData.map(enrollment => ({
       expander: {
-        lastModified: enrollment.manualEnrollment ? formatDate(enrollment.manualEnrollment.timeStamp) : 'N/A',
+        lastModified: enrollment.manualEnrollment ? enrollment.manualEnrollment.timeStamp : 'N/A',
         lastModifiedBy: enrollment.manualEnrollment && enrollment.manualEnrollment.enrolledBy ? enrollment.manualEnrollment.enrolledBy : 'N/A',
         reason: enrollment.manualEnrollment && enrollment.manualEnrollment.reason ? enrollment.manualEnrollment.reason : 'N/A',
       },
-      courseId: <a href={`${getConfig().LMS_BASE_URL}/courses/${enrollment.courseId}`} rel="noopener noreferrer" target="_blank" className="word_break">{enrollment.courseId}</a>,
+      courseId: enrollment.courseId,
       courseName: enrollment.courseName,
-      courseStart: formatDate(enrollment.courseStart),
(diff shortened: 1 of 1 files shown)


azanbinzahid:
jenkins run

src/users/enrollments/v2/Enrollments.jsx
```diff
@@ -143,23 +143,23 @@ export default function Enrollments({
         id: 'expander',
         Cell: rowExpandHandler, // Use Cell to render an expander for each row.
       },
-      {
-        Header: 'Course Run ID', accessor: 'courseId', sortable: true,
+      { // eslint-disable-next-line react/prop-types
+        Header: 'Course Run ID', accessor: 'courseId', Cell: ({ value }) => <a href={`${getConfig().LMS_BASE_URL}/courses/${value}`} rel="noopener noreferrer" target="_blank" className="word_break">{value}</a>, sortable: true,
```
DawoudSheraz, line 147: Does this issue come up only with the <a> tag?

DawoudSheraz (commented):
Not tested on local but on the surface, the changes look good. Thanks for including the references. We might need some unit tests to ensure our sorting behavior now works as expected(but can be done in a follow-up PR)

Ali-D-Akbar (approved):

ansabgillani (approved):

src/users/enrollments/v2/Enrollments.jsx
DawoudSheraz: @azanbinzahid ☝🏽

DawoudSheraz (approved):

src/users/enrollments/v2/Enrollments.jsx
azanbinzahid: This is a known bug with `react-table` and `prop-types` and Its only coming with `a` tag though.

PR #172, opened 2021-11-02
feat: add default redirect to v2

azanbinzahid:
## Description
This PR adds default redirect to v2 on first load by leveraging [session storage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage). Please note that there will be no redirect on subsequent switching or page loads until session storage expires i.e. tab or browser closes. 

## Linked Ticket
[PROD-2554](https://openedx.atlassian.net/browse/PROD-2554)

## To Test:
- Redirect to v2 on first load
- No redirect on toggle switch
- State reset on tab / browser close
- Behavior on different browsers

### DIFF
--- a/src/supportHeader/ToggleVersion.jsx
+++ b/src/supportHeader/ToggleVersion.jsx
@@ -1,4 +1,4 @@
-import React, { useContext } from 'react';
+import React, { useContext, useEffect } from 'react';
 import { Form } from '@edx/paragon';
 import { AppContext } from '@edx/frontend-platform/react';
 import { TAB_PATH_MAP } from '../SupportToolsTab/constants';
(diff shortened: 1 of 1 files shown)


DawoudSheraz:
An issue I faced while testing it out on local:
- Visit v2 URL first
- Set New Experience toggle off to go to the old experience
- When the old experience page loads, it is redirected to v2

This only happens once, but still an issue

azanbinzahid:
> An issue I faced while testing it out on local:
> 
>     * Visit v2 URL first
> 
>     * Set New Experience toggle off to go to the old experience
> 
>     * When the old experience page loads, it is redirected to v2
> 
> 
> This only happens once, but still an issue

The logic has been updated to cover above scenario.

DawoudSheraz (approved):

Ali-D-Akbar (approved):
Checked on local. Works fine. 👍

PR #175, opened 2021-11-04
fix: remove tab unmount for main tab

DawoudSheraz:
### Description
- Remove unmountOnExit from support tools tab so that the main tool does not unmount when the tab is switched. 
- Add useLayoutEffect in User and FBE Index pages responsible for updating url with appropriate query params when the tab is switched(assuming user made a search result on 1 tool, navigated to other, and then returned to original tool)

Ref: https://react-bootstrap.github.io/components/tabs/#tabs-props

### DIFF
--- a/src/FeatureBasedEnrollments/FeatureBasedEnrollmentIndexPage.jsx
+++ b/src/FeatureBasedEnrollments/FeatureBasedEnrollmentIndexPage.jsx
@@ -5,6 +5,7 @@ import React, {
   useCallback,
   useContext,
   useState,
+  useLayoutEffect,
 } from 'react';
 import PropTypes from 'prop-types';
 import { Input, Button } from '@edx/paragon';
@@ -13,6 +14,7 @@ import UserMessagesContext from '../userMessages/UserMessagesContext';
 import AlertList from '../userMessages/AlertList';
 import { isValidCourseID } from '../utils';
 import FeatureBasedEnrollment from './FeatureBasedEnrollment';
+import { FEATURE_BASED_ENROLLMENT_TAB, TAB_PATH_MAP } from '../SupportToolsTab/constants';
 
 export default function FeatureBasedEnrollmentIndexPage({ location }) {
   const params = new Map(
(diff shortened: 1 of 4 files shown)


src/FeatureBasedEnrollments/FeatureBasedEnrollmentIndexPage.jsx
```diff
@@ -82,6 +84,15 @@ export default function FeatureBasedEnrollmentIndexPage({ location }) {
     }
   }, []);
 
+  // To change the url with appropriate query param if query param info is not present in URL
+  useLayoutEffect(() => {
+    if (searchValue
+        && location.pathname.indexOf(TAB_PATH_MAP[FEATURE_BASED_ENROLLMENT_TAB]) !== -1
+        && !params.get('course_id')) {
+      pushHistoryIfChanged(`/v2/feature_based_enrollments/?course_id=${searchValue}`);
```
azanbinzahid: We can use `TAB_PATH_MAP` here as well.

src/users/v2/UserPage.jsx
```diff
@@ -119,6 +119,19 @@ export default function UserPage({ location }) {
...
 
+  // To change the url with appropriate query param if query param info is not present in URL
+  useLayoutEffect(() => {
+    if (userIdentifier
+      && location.pathname.indexOf(TAB_PATH_MAP[LEARNER_INFO_TAB]) !== -1
+      && !(params.get('email') || params.get('username'))) {
+      if (isEmail(userIdentifier)) {
+        pushHistoryIfChanged(`${TAB_PATH_MAP['learner-information']}/?email=${userIdentifier}`);
+      } else if (isValidUsername(userIdentifier)) {
+        pushHistoryIfChanged(`${TAB_PATH_MAP['learner-information']}/?username=${userIdentifier}`);
```
azanbinzahid: Curious about why its not starting from `/v2` but the other one is?

azanbinzahid (changes requested):

src/users/v2/UserPage.jsx
DawoudSheraz: the value of the dict already has v2 https://github.com/edx/frontend-app-support-tools/blob/master/src/SupportToolsTab/constants.js

Ali-D-Akbar (approved):

azanbinzahid (approved):

PR #176, opened 2021-11-05
fix: recover entitlement reissue button condition

Ali-D-Akbar:
[PROD-2560](https://openedx.atlassian.net/browse/PROD-2560)

After a discussion with Revenue team about if we should keep allowing reissue on an expired entitlement or not. It was decided that if an entitlement is used irrespective of it being expired, the Learner Support User is allowed to reissue that entitlement.

### DIFF
--- a/src/users/entitlements/v2/Entitlements.jsx
+++ b/src/users/entitlements/v2/Entitlements.jsx
@@ -107,8 +107,8 @@ export default function Entitlements({
                 setUserEntitlement(entitlement);
                 setFormType(REISSUE);
               }}
-              // disable if entitlement has no associated course run or the entitlement is expired
-              disabled={Boolean(!entitlement.enrollmentCourseRun || entitlement.expiredAt)}
+              // disable if entitlement has no associated course run
+              disabled={Boolean(!entitlement.enrollmentCourseRun)}
               className="small"
             >
               Reissue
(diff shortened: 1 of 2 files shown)


src/users/entitlements/v2/Entitlements.test.jsx
```diff
@@ -132,11 +132,6 @@ describe('Entitlements V2 Listing', () => {
 
   describe('Reissue entitlement button', () => {
     it('Enabled Reissue entitlement button', async () => {
-      entitlementsData.results[0] = { ...entitlementsData.results[0], expiredAt: null };
-      jest.spyOn(api, 'getEntitlements').mockImplementationOnce(() => Promise.resolve(entitlementsData));
-      wrapper = mount(<EntitlementsPageWrapper {...props} />);
-      await waitForComponentToPaint(wrapper);
-
```
Ali-D-Akbar, line 139: The tests were passing irrespective of these lines. Can safely be removed.

DawoudSheraz:
We might need a unit test or update in a unit test to cater the coverage issue.

src/users/entitlements/v2/Entitlements.jsx
```diff
@@ -108,7 +108,7 @@ export default function Entitlements({
                 setFormType(REISSUE);
               }}
               // disable if entitlement has no associated course run or the entitlement is expired
```
azanbinzahid: Please update comment as well if needed.

DawoudSheraz: +1 to that

DawoudSheraz (approved):

src/users/entitlements/v2/Entitlements.jsx
Ali-D-Akbar: Done.

Ali-D-Akbar: Done.

azanbinzahid (approved):

PR #198, opened 2021-11-22
feat: use enrollment date as default sort column

Ali-D-Akbar:
[PROD-2575](https://openedx.atlassian.net/browse/PROD-2575)
Use Enrollment Date as an initial sort column in Enrollment table to show the most recent enrolled course at the top of the list by default.

### DIFF
--- a/src/components/Table.jsx
+++ b/src/components/Table.jsx
@@ -5,7 +5,7 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faSortDown, faSortUp } from '@fortawesome/free-solid-svg-icons';
 
 export default function Table({
-  columns, data, renderRowSubComponent, styleName, isResponsive,
+  columns, data, renderRowSubComponent, styleName, isResponsive, defaultSortColumn,
 }) {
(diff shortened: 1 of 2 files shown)


src/components/Table.jsx
```diff
@@ -18,6 +18,12 @@ export default function Table({
     {
       columns,
       data,
+      initialState: {
+        sortBy: [{
+          id: 'defaultSortColumn',
+          desc: true,
+        }],
+      },
```
DawoudSheraz: Instead of hard-coding default sort to be desc, the component using the TableV2 should decide if the default sort is needed and in what direction.

DawoudSheraz (changes requested):
Let the component utilizing TableV2 decide if the default sort is needed in a particular direction. The hard-coding of default sort is not a scalable solution. This is also causing the test failures.

DawoudSheraz (approved):
Please uncommit package-lock.json before merge.

ansabgillani (approved):

azanbinzahid (approved):

PR #202, opened 2021-12-20
feat: add search by lms user id

azanbinzahid:
## Description
This PR adds ability to search by LMS User ID

## Linked Ticket
[PROD-2585](https://openedx.atlassian.net/browse/PROD-2585)

## SS
image

### DIFF
--- a/src/userMessages/messages.js
+++ b/src/userMessages/messages.js
@@ -1,4 +1,5 @@
-export const USER_IDENTIFIER_INVALID_ERROR = 'The searched username or email is invalid. Please correct the username or email and try again.';
+export const USER_IDENTIFIER_INVALID_ERROR = 'The searched username, email or lms user id is invalid. Please correct the username, email or lms user id and try again.';
 export const USERNAME_IDENTIFIER_NOT_FOUND_ERROR = 'We couldn\'t find a user with the username "{identifier}".';
 export const USER_EMAIL_IDENTIFIER_NOT_FOUND_ERROR = 'We couldn\'t find a user with the email "{identifier}".';
+export const LMS_USER_ID_IDENTIFIER_NOT_FOUND_ERROR = 'We couldn\'t find a user with the LMS User ID "{identifier}".';
 export const UNKNOWN_API_ERROR = "There was an error loading this user's data. Check the JavaScript console for detailed errors.";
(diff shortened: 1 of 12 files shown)


src/userMessages/messages.js
```diff
@@ -1,4 +1,5 @@
-export const USER_IDENTIFIER_INVALID_ERROR = 'The searched username or email is invalid. Please correct the username or email and try again.';
+export const USER_IDENTIFIER_INVALID_ERROR = 'The searched username, email or LMS ID is invalid. Please correct the username or email and try again.';
```
DawoudSheraz: nit: to update the 2nd part of the message to include LMS id

DawoudSheraz (approved):
Please add appropriate screenshots in PR description. Otherwise, the changes are good to go.

Ali-D-Akbar (approved):
Just a suggestion. LMS ID might not be as self explanatory. If we can write `LMS user ID` instead, that'd really help.
Looks good to me otherwise. 👍

src/users/UserPage.jsx
```diff
@@ -48,16 +48,28 @@ export default function UserPage({ location }) {
     }
   }
 
+  function getUpdatedURL(value) {
+    const updatedHistory = `/users/?PARAM_NAME=${value}`;
+    let identifierType = '';
+
+    if (isEmail(value)) {
```
ansabgillani, line 55: You can use switch or key, value identifier over here rather than if-else if branches.

src/users/data/api.js
```diff
@@ -104,10 +106,17 @@ export async function getUser(userIdentifier) {
     // never do this in a customer-facing app.
     // eslint-disable-next-line no-console
     console.log(JSON.parse(error.customAttributes.httpErrorResponseData));
-    const notFoundErrorText = (isEmail(userIdentifier)
-      ? messages.USER_EMAIL_IDENTIFIER_NOT_FOUND_ERROR
-      : messages.USERNAME_IDENTIFIER_NOT_FOUND_ERROR
-    ).replace('{identifier}', userIdentifier);
+
+    let notFoundErrorText = null;
+    if (isEmail(userIdentifier)) {
```
ansabgillani, line 111: Same as above.

ansabgillani (approved):

ansabgillani:
A few code quality improvements can be made. It's not a blocker though.

src/users/UserPage.jsx
azanbinzahid: The 3 functions return a boolean value, can you suggest a way in which it can be catered via switch?

ansabgillani: `def operation(operator, x, y): 
 return {
   'add': lambda: x+y,
   'sub': lambda: x-y,
   'mul': lambda: x*y,
   'div': lambda: x/y
 }.get(operator, lambda: None)()`
 
 This is something that I had in mind. Nevermind. though, it's equally better.

ansabgillani (approved):

PR #208, opened 2022-01-12
feat: add verified name to user summary

ansabgillani:
### Description
This PR is related to the request to add a verified name in support tools MFE on the account information section. The verified name is defined in edx-name-affirmation's VerifiedName model.  See https://openedx.atlassian.net/wiki/spaces/PT/pages/2629928102/Name+Affirmation+Re-imagine+IDV on how it looks. 
This PR adds a user field Verified Name in User Summary that returns the latest approved verified name.

### Testing Instructions
[Backend PR](https://github.com/edx/edx-platform/pull/29624)

Screenshot:
Screenshot 2022-01-12 at 5 06 12 PM

### DIFF
--- a/src/users/v2/UserSummary.jsx
+++ b/src/users/v2/UserSummary.jsx
@@ -27,6 +27,11 @@ export default function UserSummary({
                 </tr>
 
                 <tr>
+                  <th>Verified Name</th>
+                  <td>{userData.verifiedName}</td>
+                </tr>
+
+                <tr>
                   <th>Username</th>
                   <td>{userData.username}</td>
                 </tr>
@@ -102,6 +107,7 @@ export default function UserSummary({
 UserSummary.propTypes = {
   userData: PropTypes.shape({
     name: PropTypes.string,
+    verifiedName: PropTypes.string,
     username: PropTypes.string,
     id: PropTypes.number,
     email: PropTypes.string,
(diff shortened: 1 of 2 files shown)


src/users/v2/UserSummary.jsx
```diff
@@ -26,6 +26,11 @@ export default function UserSummary({
                   <td>{userData.name}</td>
                 </tr>
 
+                <tr>
+                  <th>Verified Name</th>
+                  <td>{userData.verifiedName ? userData.verifiedName : 'N/A'}</td>
```
DawoudSheraz: nit: we can keep the row empty instead of showing N/A. Or rename N/A to unavailable.

src/users/v2/UserSummary.test.jsx
```diff
@@ -58,6 +58,10 @@ describe('User Summary Component Tests', () => {
           header: 'Full Name',
           value: 'edx',
         },
+        {
+          header: 'Verified Name',
+          value: 'N/A',
+        },
```
DawoudSheraz: Relating to above question, if if condition is kept, we should add another small test case to showcase the correct value is displayed if verified name is present.

DawoudSheraz (approved):
Nits to address before merge.

src/users/v2/UserSummary.jsx
ansabgillani: Keeping it empty makes more sense as name field is also empty if there is no name for a user.

azanbinzahid (approved):

azanbinzahid:
Please add SS

Ali-D-Akbar (approved):

PR #214, opened 2022-01-25
feat: display error response on password reset

AliAdnanSohail:
[show reset password api response on dialog](https://openedx.atlassian.net/browse/PROD-2600)
Screen Shot 2022-01-31 at 8 23 16 PM

### DIFF
--- a/src/users/account-actions/v2/ResetPassword.jsx
+++ b/src/users/account-actions/v2/ResetPassword.jsx
@@ -9,11 +9,40 @@ export default function ResetPassword({
   changeHandler,
 }) {
   const [resetPasswordModalIsOpen, setResetPasswordModalIsOpen] = useState(false);
+  const [errorMessage, setErrorMessage] = useState(null);
 
-  const resetPassword = () => {
-    postResetPassword(email);
-    changeHandler();
+  const resetPassword = async () => {
+    const resp = await postResetPassword(email);
+    if (resp.errors) {
+      setErrorMessage(resp.errors[0].text || 'Something went wrong. Please try again later!');
+    } else {
+      changeHandler();
+    }
   };
+
+  const closeResetPasswordModal = () => {
+    setResetPasswordModalIsOpen(false);
+    setErrorMessage(null);
+  };
+
+  const modalBody = (
+    errorMessage ? <Alert variant="danger">{errorMessage}</Alert>
+      : (
+        <div>
+          <Alert variant="warning">
+            <FormattedMessage
+              id="supportTools.accountActions.resetPassword"
+              tagName="p"
+              description="Password reset email instructions prompt"
+              defaultMessage="We will send a message with password recovery instructions to the email address {email}. Do you wish to proceed?"
+              values={{
+                email: <strong>{email}</strong>,
+              }}
+            />
+          </Alert>
+        </div>
+      )
+  );
 
   return (
     <div>
@@ -28,33 +57,20 @@ export default function ResetPassword({
       <Modal
         open={resetPasswordModalIsOpen}
         id="user-account-reset-password"
-        buttons={[
-          <Button
-            variant="danger"
-            onClick={resetPassword}
-          >
-            Confirm
-          </Button>,
+        buttons={[errorMessage ? (<></>)
+          : (
+            <Button
+              variant="danger"
+              onClick={resetPassword}
+            >
+              Confirm
+            </Button>
+          ),
         ]}
-        onClose={() => setResetPasswordModalIsOpen(false)}
+        onClose={closeResetPasswordModal}
         dialogClassName="modal-lg modal-dialog-centered justify-content-center"
         title="Reset Password"
-        body={(
-          <div>
-            <Alert variant="warning">
-              <FormattedMessage
-                id="supportTools.accountActions.resetPassword"
-                tagName="p"
-                description="Password reset email instructions prompt"
-                defaultMessage="We will send a message with password recovery instructions to the email address {email}.
-                Do you wish to proceed?"
-                values={{
-                  email: <strong>{email}</strong>,
-                }}
-              />
-            </Alert>
-          </div>
-          )}
+        body={modalBody}
       />
     </div>
   );
--- a/src/users/account-actions/v2/ResetPassword.test.jsx
+++ b/src/users/account-actions/v2/ResetPassword.test.jsx
@@ -4,6 +4,7 @@ import { IntlProvider } from '@edx/frontend-platform/i18n';
 import * as api from '../../data/api';
 import ResetPassword from './ResetPassword';
 import UserSummaryData from '../../data/test/userSummary';
+import { waitForComponentToPaint } from '../../../setupTest';
 
 const ResetPasswordWrapper = (props) => (
   <IntlProvider locale="en">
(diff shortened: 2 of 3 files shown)


src/users/account-actions/v2/ResetPassword.jsx
```diff
@@ -56,6 +64,26 @@ export default function ResetPassword({
           </div>
           )}
       />
+      <Modal
```
azanbinzahid: I believe we can use the same model just render different body based on error status, will save a few lines of code.

src/users/data/api.js
```diff
@@ -507,10 +507,24 @@ export async function postTogglePasswordStatus(user, comment) {
...
+      AppUrls.getResetPasswordUrl(), `email_from_support_tools=${email}`,
+    );
+    return data;
+  } catch (error) {
+    return {
+      errors: [
+        {
+          code: null,
+          dismissible: true,
+          text: error.response.data,
```
azanbinzahid: What kind of errors can we expect from this API? Will there be any case in which `error.response.data` will not be there in an error say `500`?

src/users/account-actions/v2/ResetPassword.test.jsx
```diff
@@ -49,11 +50,46 @@ describe('Reset Password Component Tests', () => {
...
     resetPasswordModal.find('button.btn-link').simulate('click');
     resetPasswordModal = wrapper.find('Modal#user-account-reset-password');
     expect(resetPasswordModal.prop('open')).toEqual(false);
+
+    const resetPasswordErrorModal = wrapper.find('Modal#user-account-reset-password-error');
+    expect(resetPasswordErrorModal.prop('open')).toEqual(false);
+    mockApiCall.mockRestore();
+  });
+
+  it('Display Error on Reset Password Modal', async () => {
```
azanbinzahid, line 62: You may need to test with all possible (or a few more) kind of errors.

src/users/account-actions/v2/ResetPassword.jsx
```diff
@@ -56,6 +64,26 @@ export default function ResetPassword({
...
+        title="Reset Password Error"
+        body={(
+          <div>
+            <Alert variant="warning">
+              <FormattedMessage
+                id="supportTools.accountActions.resetPasswordError"
+                tagName="p"
+                description="Password reset email errors prompt"
+                defaultMessage="{errorMessage}"
+                values={{ errorMessage }}
```
azanbinzahid: Not sure what's the difference b/w the two `{{testString}}` or `"{testString}"`

azanbinzahid (changes requested):

src/users/account-actions/v2/ResetPassword.jsx
AliAdnanSohail: `defaultMessage` attribute should be translatable so we could not pass dynamic value in this. To pass dynamic value we have to interpolate that dynamic value and then pass that value in `values` attribute

src/users/account-actions/v2/ResetPassword.test.jsx
AliAdnanSohail: Actually I've catch all type of errors which will be returned in `errors` object in same way. So we don't need to test different type of errors because only string will be different, everything else will be same.

src/users/account-actions/v2/ResetPassword.jsx
AliAdnanSohail: Actually same modal contains some extra buttons and their click handlers. So either we need to add conditions to display or hide them, or we can handle this in separate modal. So I choose the second option

src/users/data/api.js
AliAdnanSohail: Yes we can add a check if there is no error description and error.response.data is empty then we can display some generic message e.g `Something went wrong` may be

Ali-D-Akbar (commented):
Overall the changes look good. Just add some tests to cover all the checks. There is some drop in the coverage.

src/users/account-actions/v2/ResetPassword.jsx
DawoudSheraz: +1 to what Azan said. The button/click handlers will only be closing the modal and not doing anything special. We can also hide buttons under certain conditions too.

DawoudSheraz (commented):
Most concerns have been raised by Azan already.

src/users/account-actions/v2/ResetPassword.jsx
azanbinzahid: I believe we can avoid using formatted messages and use simple string. Please see relevant usage [here](https://github.com/openedx/frontend-app-support-tools/blob/76d3b5c5756bd0d9842817b247fe9a387d2cbd2a/src/users/enrollments/v2/Certificates.jsx).

src/users/account-actions/v2/ResetPassword.jsx
```diff
@@ -9,12 +9,44 @@ export default function ResetPassword({
   changeHandler,
 }) {
   const [resetPasswordModalIsOpen, setResetPasswordModalIsOpen] = useState(false);
+  const [errorMessage, setErrorMessage] = useState(null);
+  const [resetPasswordError, setResetPasswordError] = useState(false);
```
azanbinzahid: We can use one state variable for both.

src/users/account-actions/v2/ResetPassword.jsx
```diff
@@ -28,33 +60,20 @@ export default function ResetPassword({
...
         open={resetPasswordModalIsOpen}
         id="user-account-reset-password"
-        buttons={[
-          <Button
-            variant="danger"
-            onClick={resetPassword}
-          >
-            Confirm
-          </Button>,
+        buttons={[resetPasswordError ? ('')
```
azanbinzahid: you may return `null` or` <></>` as well.

azanbinzahid (approved):

Ali-D-Akbar (approved):

src/users/account-actions/v2/ResetPassword.jsx
DawoudSheraz: Yes, we can check the existence of the error from error message.

src/users/account-actions/v2/ResetPassword.jsx
DawoudSheraz: +1

src/users/account-actions/v2/ResetPassword.jsx
```diff
@@ -9,12 +9,44 @@ export default function ResetPassword({
...
-  const resetPassword = () => {
-    postResetPassword(email);
-    changeHandler();
+  const resetPassword = async () => {
+    const resp = await postResetPassword(email);
+    if (resp.errors) {
+      setResetPasswordError(true);
+      setErrorMessage(resp.errors[0].text || 'Something went wrong. Please try again later!');
+    } else {
+      changeHandler();
```
DawoudSheraz, line 19: personally, I think we can remove this changeHandler and display some messages, such as a request is submitted. We can do this in a follow-up ticket and PR. Just flagging this now that it is in the works, we can go with fixing all behavioral inconsistencies. The page reloadiing after successful submission seems a bit weird now.

DawoudSheraz (approved):

src/users/account-actions/v2/ResetPassword.jsx
AliAdnanSohail: @DawoudSheraz Actually its coming from props, and if we don't need the functionality of this specific method then we can simply close the modal and update message in modal

DawoudSheraz: Yes, something like that. We can remove submit button and only display a success/error message based on the response. Closing modal will be left to the user.

AliAdnanSohail: If you want me to handle in this ticket. I can do it here as well

DawoudSheraz: no, lets have a dedicated ticket and PR for that.

PR #216, opened 2022-01-26
[MST-1135] Remove IDV and add verified name panel

bseverino:
https://openedx.atlassian.net/browse/MST-1135

From the V2 experience, remove the ID verification panel and add one with information about the user's verified name(s).

Screen Shot 2022-01-27 at 9 18 45 AM

Screen Shot 2022-01-27 at 9 19 48 AM


schenedx (approved):
👍

PR #217, opened 2022-02-08
feat: Display verified-in and current-status data on onboarding status

AliAdnanSohail:
[PROD-2584](https://openedx.atlassian.net/browse/PROD-2584)

### Description

- Show the current Onboarding status and the actual verified-in course status details on v2 user summary
- Update some failing tests due to async updates after the test completion

Screen Shot 2022-02-09 at 2 53 34 PM

### DIFF
--- a/src/users/data/api.js
+++ b/src/users/data/api.js
@@ -365,7 +365,18 @@ export async function getV2OnboardingStatus(username) {
     );
     return data;
   } catch (error) {
-    return defaultResponse;
+    let errorText = 'Error while fetching data';
+
+    try {
+      if (error.customAttributes?.httpErrorResponseData) {
+        errorText = JSON.parse(error.customAttributes.httpErrorResponseData);
+      }
+    } catch (e) {
+      // In case there is something wrong with the response, use the default
+      // error message
+    }
+
+    return { ...defaultResponse, error: errorText };
   }
 }
 
--- a/src/users/data/api.js
+++ b/src/users/data/api.js
@@ -354,6 +354,21 @@ export async function getOnboardingStatus(enrollments, username) {
   }
 }
 
+export async function getV2OnboardingStatus(username) {
+  const defaultResponse = {
+    verified_in: null,
+    current_status: null,
+  };
+  try {
+    const { data } = await getAuthenticatedHttpClient().get(
+      AppUrls.getV2OnboardingStatusUrl(username),
+    );
+    return data;
+  } catch (error) {
+    return defaultResponse;
+  }
+}
+
 export async function getAllUserData(userIdentifier) {
   const errors = [];
   let user = null;
--- a/src/users/data/test/onboardingStatus.js
+++ b/src/users/data/test/onboardingStatus.js
@@ -7,4 +7,29 @@ const onboardingStatusData = {
   reviewRequirementsUrl: null,
 };
 
-export default onboardingStatusData;
+const v2OnboardingStatusData = {
+  verifiedIn: {
+    onboardingStatus: 'verified',
+    onboardingLink: '/courses/course-v1:test-course/jump_to/block-v1:test-course+type@sequential+block@12345678',
+    expirationDate: '2022-09-07T23:59:59+00:00',
+    onboardingPastDue: false,
+    onboardingReleaseDate: '2020-09-03T20:00:00+00:00',
+    reviewRequirementsUrl: 'https://support.edx.org/hc/en-us/sections/12345678-Proctortrack',
+    courseId: 'course-v1:test-course',
+    enrollmentDate: '2020-07-03T08:57:50Z',
+    instructorDashboardLink: '/courses/course-v1:test-course/instructor#view-special_exams',
+  },
+  currentStatus: {
+    onboardingStatus: 'verified',
+    onboardingLink: '/courses/course-v1:test-course/jump_to/block-v1:test-course+type@sequential+block@12345678',
+    expirationDate: '2022-09-07T23:59:59+00:00',
+    onboardingPastDue: false,
+    onboardingReleaseDate: '2020-09-03T20:00:00+00:00',
+    reviewRequirementsUrl: 'https://support.edx.org/hc/en-us/sections/12345678-Proctortrack',
+    courseId: 'course-v1:test-course',
+    enrollmentDate: '2020-07-03T08:57:50Z',
+    instructorDashboardLink: '/courses/course-v1:test-course/instructor#view-special_exams',
+  },
+};
+
+export { onboardingStatusData, v2OnboardingStatusData };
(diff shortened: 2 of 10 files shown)


src/users/data/test/onboardingStatus.js
```diff
@@ -5,6 +5,28 @@ const onboardingStatusData = {
...
+    onboardingStatus: 'verified',
+    onboardingLink: '/courses/course-v1:test-course/jump_to/block-v1:test-course+type@sequential+block@12345678',
+    expirationDate: '2022-09-07T23:59:59+00:00',
+    onboardingPastDue: false,
+    onboardingReleaseDate: '2020-09-03T20:00:00+00:00',
+    reviewRequirementsUrl: 'https://support.edx.org/hc/en-us/sections/12345678-Proctortrack',
+    courseId: 'course-v1:test-course',
+    enrollmentDate: '2020-07-03T08:57:50Z',
+    instructorDashboardLink: '/courses/course-v1:test-course/instructor#view-special_exams',
+  },
```
DawoudSheraz, line 32: Shouldn't this be a separate data variable?

src/users/v2/OnboardingStatus.jsx
```diff
@@ -30,33 +34,45 @@ export default function OnboardingStatus({
...
     <div className="flex-column p-4 m-3 card">
       <h3>Proctoring Information</h3>
-      {onboardingData ? (
-        <TableV2
-          id="proctoring-data"
-          data={proctoringData}
-          columns={proctoringColumns}
-          styleName="idv-table"
-        />
+      { onboardingData ? (
```
DawoudSheraz, line 55: how are we not changing API call here? We introduced a new endpoint in lms support app and would be using that, isn't it the case?
cc: @azanbinzahid

DawoudSheraz (changes requested):
API call updates are missing.

src/users/v2/OnboardingStatus.jsx
DawoudSheraz: https://github.com/openedx/edx-platform/pull/29607 this is the PR/API I am referring to.

src/users/data/test/onboardingStatus.js
azanbinzahid: +1
Please use a different variable with mock data.

src/users/v2/OnboardingStatus.jsx
azanbinzahid: This will be updated, it was done to fast-forward the implementation and must've been mentioned in PR description.

azanbinzahid (changes requested):
I think we can now update/add call with new endpoint and complete this PR. Meanwhile, please keep this PR as draft.

src/users/data/api.js
```diff
@@ -354,6 +354,21 @@ export async function getOnboardingStatus(enrollments, username) {
...
+    verified_in: null,
+    current_status: null,
+  };
+  try {
+    const { data } = await getAuthenticatedHttpClient().get(
+      AppUrls.getV2OnboardingStatusUrl(username),
+    );
+    return data;
+  } catch (error) {
+    return defaultResponse;
```
azanbinzahid: may be we should return/display some kind of error as well.

src/users/entitlements/v2/Entitlements.test.jsx
```diff
@@ -109,20 +109,22 @@ describe('Entitlements V2 Listing', () => {
   });
 
   describe('Expire Entitlement button', () => {
-    it('Disabled Expire entitlement button', () => {
+    it('Disabled Expire entitlement button', async () => {
```
azanbinzahid, line 112: Please add explanation as to why it was[n't] working.

src/users/v2/OnboardingStatus.jsx
```diff
@@ -30,33 +32,45 @@ export default function OnboardingStatus({
...
-  const proctoringData = useMemo(() => {
-    if (onboardingData === null || onboardingData.length === 0) {
-      return [];
-    }
-    return [onboardingData].map(result => ({
-      status: result.onboardingStatus ? titleCase(result.onboardingStatus) : 'Not Started',
+  function formatData(data) {
+    return [data].map(result => ({
+      courseId: result.courseId || 'No Course',
+      status: result.onboardingStatus ? titleCase(result.onboardingStatus) : 'See Dashboard',
```
azanbinzahid: Please add a comment on why we are redirecting to dashboard.

src/users/v2/OnboardingStatus.jsx
```diff
@@ -30,33 +32,45 @@ export default function OnboardingStatus({
...
+          ) : <div className="no-record-text" id="verified-in-no-data">No Record Found</div>}
+          <h4>Current Status</h4>
+          { onboardingData.currentStatus ? (
+            <TableV2
+              id="current-status-data"
+              data={formatData(onboardingData.currentStatus)}
+              columns={proctoringColumns}
+              styleName="idv-table"
+            />
+          ) : <div className="no-record-text" id="current-status-no-data">No Record Found</div>}
```
azanbinzahid: error can be displayed here

src/users/data/api.js
DawoudSheraz: Yup, true.

src/users/v2/OnboardingStatus.jsx
DawoudSheraz: The error returned in API response?

src/users/entitlements/v2/Entitlements.test.jsx
DawoudSheraz: Yeah, this was a weird one. When running a few test cases individually, the Entitlement component tests were constantly causing giving logs errors that async updates were being made. The DropDown component, which shows the possible actions on entitlement after click, was having its state updated internally after the test execution was complete. That's why I added wait after click to allow DropDown updates to complete before moving ahead.

src/users/v2/OnboardingStatus.jsx
DawoudSheraz: Not sure why @AliAdnanSohail did that. We can maybe add a default status here instead of See Dashboard text.

azanbinzahid: Actually I suggested this and having  comment in code will mitigate the confusion. There are certain status in db for which the api gives null/undefined value but they are visible on instructor dashboard.

src/users/v2/OnboardingStatus.jsx
azanbinzahid: Yes

src/users/v2/OnboardingStatus.jsx
DawoudSheraz: Right, makes sense.

azanbinzahid (approved):

DawoudSheraz (approved):

PR #220, opened 2022-02-15
fix: Responsiveness of Enrollment and Entitlement V2 table

AliAdnanSohail:
[entitlement action hiding behind enrollment button for a single entitlement](https://openedx.atlassian.net/browse/PROD-2571)

Description: In case of single row drop down was hiding behind its below component because our table is responsive and add scroll if table content exceed. We need this responsive behavior so we have handled this edge case of single row manually by increasing size of table div, so drop can properly visible

Screen Shot 2022-02-17 at 7 51 27 PM

### DIFF
--- a/src/overrides.scss
+++ b/src/overrides.scss
@@ -164,3 +164,7 @@ $darkCyan: #00262b;
 .background-light-gray {
   background: lightgrey;
 }
+
+.mb-60 {
+  margin-bottom: 60px;
+}
--- a/src/users/enrollments/v2/Enrollments.jsx
+++ b/src/users/enrollments/v2/Enrollments.jsx
@@ -281,7 +281,7 @@ export default function Enrollments({
             columns={columns}
             data={tableData}
             renderRowSubComponent={renderRowSubComponent}
-            styleName="custom-table"
+            styleName={tableData.length === 1 ? 'custom-table mb-60' : 'custom-table'}
             defaultSortColumn={defaultSortColumn}
           />
         )
--- a/src/users/entitlements/v2/Entitlements.jsx
+++ b/src/users/entitlements/v2/Entitlements.jsx
@@ -268,7 +268,7 @@ export default function Entitlements({
             columns={columns}
             data={tableData}
             renderRowSubComponent={renderRowSubComponent}
-            styleName="custom-table"
+            styleName={tableData.length === 1 ? 'custom-table mb-60' : 'custom-table'}
           />
         )}
 


src/users/enrollments/v2/Enrollments.jsx
```diff
@@ -281,7 +281,7 @@ export default function Enrollments({
             columns={columns}
             data={tableData}
             renderRowSubComponent={renderRowSubComponent}
-            styleName="custom-table"
+            styleName={tableData.length === 1 ? 'custom-table mb-60' : 'custom-table'}
```
azanbinzahid, line 284: Can be reduced to `tableData.length` only.

src/users/entitlements/v2/Entitlements.jsx
```diff
@@ -268,7 +268,7 @@ export default function Entitlements({
             columns={columns}
             data={tableData}
             renderRowSubComponent={renderRowSubComponent}
-            styleName="custom-table"
+            styleName={tableData.length === 1 ? 'custom-table mb-60' : 'custom-table'}
```
azanbinzahid, line 271: same as above ^

azanbinzahid:
Change looks good, make sure to test UI locally (for 0, 1 and n rows) before merging.

src/users/enrollments/v2/Enrollments.jsx
AliAdnanSohail: @azanbinzahid - Actually we need to add this class only if record length is exactly equal to 1

src/overrides.scss
```diff
@@ -164,3 +164,7 @@ $darkCyan: #00262b;
 .background-light-gray {
   background: lightgrey;
 }
+
+.mb-60 {
+  margin-bottom: 60px;
+}
```
DawoudSheraz, line 170: can we use bootstrap's mb-{val} class instead of writing our own? 
https://getbootstrap.com/docs/4.0/utilities/spacing/#notation

DawoudSheraz (approved):
- Update PR Description as to why the margin class is added
- To verify if bootstrap margin class can be used instead of custom class.

azanbinzahid (approved):

src/overrides.scss
AliAdnanSohail: @DawoudSheraz actually by default it bootstrap provide spacing till mb-5. And this is not enough which is required here. To add required we have to add custom spacing in spacer. But not sure do we need exactly 60px any where else or not. So I didn't achieve this by custom bootstrap

DawoudSheraz: ok, thanks.

PR #232, opened 2022-03-09
fix: display error message returned from API on enrollment create

AliAdnanSohail:
[Improving Feedback message for support using "Create New Enrollment" actions](https://openedx.atlassian.net/browse/PROD-2607)

Description: Before this we were not displaying exact error which is returned by API rather we were display error message to check console for details. But now we are displaying actual error which is returned by API

Screen Shot 2022-03-09 at 1 40 52 PM

### DIFF
--- a/src/users/data/api.js
+++ b/src/users/data/api.js
@@ -527,17 +527,18 @@ export async function postEnrollment({
     );
     return data;
   } catch (error) {
+    let errorMessage;
     if (error.customAttributes.httpErrorStatus === 400) {
+      errorMessage = JSON.parse(error.customAttributes.httpErrorResponseData);
       // eslint-disable-next-line no-console
-      console.log(JSON.parse(error.customAttributes.httpErrorResponseData));
+      console.log(errorMessage);
     }
     return {
       errors: [
         {
           code: null,
           dismissible: true,
-          text:
-            'There was an error creating the enrollment. Check the JavaScript console for detailed errors.',
+          text: errorMessage || 'An unexpected error occurred',
           type: 'danger',
           topic: 'createEnrollments',
         },
--- a/src/users/data/api.test.js
+++ b/src/users/data/api.test.js
@@ -815,12 +815,25 @@ describe('API', () => {
         const expectedError = {
           code: null,
           dismissible: true,
-          text:
-          'There was an error creating the enrollment. Check the JavaScript console for detailed errors.',
+          text: 'An unexpected error occurred',
           type: 'danger',
           topic: 'enrollments',
         };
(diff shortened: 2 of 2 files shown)


src/users/data/api.js
```diff
@@ -527,17 +527,19 @@ export async function postEnrollment({
     );
     return data;
   } catch (error) {
+    let error_message;
     if (error.customAttributes.httpErrorStatus === 400) {
+      error_message = JSON.parse(error.customAttributes.httpErrorResponseData);
```
azanbinzahid: do we intent to capture error other than `400`?

src/users/data/api.js
```diff
@@ -527,17 +527,19 @@ export async function postEnrollment({
     );
     return data;
   } catch (error) {
+    let error_message;
```
Ali-D-Akbar: use `errorMessage` instead.

src/users/data/api.test.js
```diff
@@ -825,6 +825,20 @@ describe('API', () => {
         expect(...response.errors).toEqual({ ...expectedError, topic: 'createEnrollments' });
       });
 
+      it('Unsuccessful enrollment create with error message', async () => {
+        const expectedError = {
+          code: null,
+          dismissible: true,
+          text:
+          'User already enrolled',
```
azanbinzahid, line 832: can they be in same line?

Ali-D-Akbar (approved):
Looks good to me.

src/users/data/api.js
AliAdnanSohail: Ideally in case of error it should return 400 but for safe side in case of any other error or unknown error it will display generic message.
`An unexpected error occurred`

azanbinzahid (approved):

src/users/data/api.js
```diff
@@ -527,17 +527,18 @@ export async function postEnrollment({
...
-      console.log(JSON.parse(error.customAttributes.httpErrorResponseData));
+      console.log(errorMessage);
     }
     return {
       errors: [
         {
           code: null,
           dismissible: true,
-          text:
-            'There was an error creating the enrollment. Check the JavaScript console for detailed errors.',
```
ansabgillani, line 540: Add the error as an Alert rather than keeping it for JS console.

ansabgillani (approved):

src/users/data/api.js
AliAdnanSohail: There are some more APIs which are displaying error on console e.g `patchEnrollment` We can create a separate task to display in alert rather than console

PR #233, opened 2022-03-09
feat: show user message on retired email search

Ali-D-Akbar:
[PROD-2521](https://openedx.atlassian.net/browse/PROD-2521)
Screenshot 2022-03-10 at 2 18 37 AM

For the retired email, we weren't able to get anything from Support Tools hence we changed the message here so that it explains why the account info associated to the email cannot be received.

### DIFF
--- a/src/users/data/api.js
+++ b/src/users/data/api.js
@@ -108,7 +108,9 @@ export async function getUser(userIdentifier) {
     console.log(JSON.parse(error.customAttributes.httpErrorResponseData));
 
     let notFoundErrorText = null;
-    if (isEmail(userIdentifier)) {
+    if (JSON.parse(error.customAttributes?.httpErrorResponseData)) {
+      notFoundErrorText = JSON.parse(error.customAttributes.httpErrorResponseData).error_msg;
+    } else if (isEmail(userIdentifier)) {
       notFoundErrorText = messages.USER_EMAIL_IDENTIFIER_NOT_FOUND_ERROR;
     } else if (isValidLMSUserID(userIdentifier)) {
       notFoundErrorText = messages.LMS_USER_ID_IDENTIFIER_NOT_FOUND_ERROR;
--- a/src/users/data/api.test.js
+++ b/src/users/data/api.test.js
@@ -8,6 +8,7 @@ import verifiedNameHistoryData from './test/verifiedNameHistory';
 import { v2OnboardingStatusData } from './test/onboardingStatus';
 import * as api from './api';
 import * as urls from './urls';
+import * as messages from '../../userMessages/messages';
 
 describe('API', () => {
   const testUsername = 'username';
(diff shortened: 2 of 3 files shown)


src/users/data/api.js
```diff
@@ -108,7 +108,16 @@ export async function getUser(userIdentifier) {
...
+    if (JSON.parse(error.customAttributes.httpErrorResponseData)) {
+      error.userError = {
+        code: null,
+        dismissible: true,
+        text: JSON.parse(error.customAttributes.httpErrorResponseData).error_msg,
+        type: 'error',
+        topic: 'general',
+      };
+      throw error;
+    } else if (isEmail(userIdentifier)) {
```
azanbinzahid: I think we should just set `notFoundErrorText` here and follow the existing flow. Error is thrown after `if/else` block.

azanbinzahid (changes requested):

src/users/data/api.js
```diff
@@ -108,7 +108,9 @@ export async function getUser(userIdentifier) {
     console.log(JSON.parse(error.customAttributes.httpErrorResponseData));
 
     let notFoundErrorText = null;
-    if (isEmail(userIdentifier)) {
+    if (JSON.parse(error.customAttributes.httpErrorResponseData)) {
+      notFoundErrorText = JSON.parse(error.customAttributes.httpErrorResponseData).error_msg;
```
azanbinzahid: One more thing, you may need to check if these attrib exist or not, just to prevent any unexpected error.

azanbinzahid (approved):

src/users/data/api.js
```diff
@@ -108,7 +108,9 @@ export async function getUser(userIdentifier) {
     console.log(JSON.parse(error.customAttributes.httpErrorResponseData));
 
     let notFoundErrorText = null;
-    if (isEmail(userIdentifier)) {
+    if (JSON.parse(error.customAttributes?.httpErrorResponseData)) {
```
DawoudSheraz, line 111: ?. what is this syntax?

src/users/data/api.test.js
```diff
@@ -451,6 +452,22 @@ describe('API', () => {
...
+        topic: 'general',
+      };
+      mockAdapter.onGet(`${userAccountApiBaseUrl}?email=${encodeURIComponent(testEmail)}`).reply(() => throwError(404, { error_msg: messages.USER_RETIRED_EMAIL_IDENTIFIER_ERROR }));
+      try {
+        await api.getUser(testEmail);
+      } catch (error) {
+        expect(error.userError).toEqual(expectedUserError);
+      }
+    });
+
```
DawoudSheraz: A bit difficult to read here. Add the comment on why the error handling is needed here.

src/users/data/api.js
Ali-D-Akbar: if `customAttributes` exists (isn't false), get `httpErrorResponseData`.

DawoudSheraz: Is this Es6 syntax? Seeing this for the first time.

DawoudSheraz (approved):
one small nit comment, otherwise good to go.

src/users/data/api.js
Ali-D-Akbar: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

src/users/data/api.test.js
Ali-D-Akbar: improved readability and added a comment too.

PR #234, opened 2022-03-10
fix: fix SSO Records in Program Inspector Production

ansabgillani:
This is a fix of SSO records in Inspectors. Previously, they were connected to saml provider type but now they are connected to program uids. 

Ticket-link: [PROD-2737](https://openedx.atlassian.net/browse/PROD-2737)


DawoudSheraz (approved):
once the coverage issue is resolved and build is green.

PR #239, opened 2022-03-29
feat: remove enrollments entitlements and license manager from old view

Ali-D-Akbar:
[PROD-2757](https://openedx.atlassian.net/browse/PROD-2757)
We've started the deprecation for the old Support Tools MFE in a number of phases. This is the first PR of the deprecation process and it removes old components like enrollment, entitlement and license manager. 

Screenshot 2022-03-30 at 12 59 14 AM


azanbinzahid:
Shouldn't we remove v2 toggle button along with it?

src/users/UserPage.jsx
```diff
@@ -37,9 +34,6 @@ export default function UserPage({ location }) {
   const [searching, setSearching] = useState(false);
```
DawoudSheraz, line 34: don't we need to delete this file too?

DawoudSheraz:
+1 to what @azanbinzahid has said. The toggle should be removed too. Then, in the follow-up PR, the URLs would need an update where all the references to `v2` would be removed. But do not merge this PR until the relevant parties have been notified.

DawoudSheraz (approved):
didn't test it out on the local. Straightforward removal.

azanbinzahid (approved):

PR #243, opened 2022-03-30
feat: remove IDV, SSO, Onboarding status and Account Info from old experience

Ali-D-Akbar:
[PROD-2760](https://openedx.atlassian.net/browse/PROD-2760)
Simply removing IDV, SSO, Onboarding status and User Summary. The current version of old experience only shows a search bar and shows no info.


DawoudSheraz (approved):
We can think of removing the old Table component in the follow-up PR too as it is highly likely it won't be used anymore after old experience removal.

azanbinzahid (approved):

PR #249, opened 2022-04-05
refactor: use v2 as default

Ali-D-Akbar:
[PROD-2762](https://openedx.atlassian.net/browse/PROD-2762)
Sets v2 as the default UI for Support Tools MFE. All the occurrences for v2 has been either removed or refactored into their respective parent directories. Reviewers are suggested to please take a good look at it so that nothing is broken and the v2 UI is set as default.


ansabgillani (approved):

DawoudSheraz (approved):
For a follow-up PR, we can use tab and mapping variables defined in src/SupportToolsTab/constants.js when replacing history urls across FBE, Program, and User pages.

PR #253, opened 2022-04-20
feat: add learner credentials

ansabgillani:
This PR introduces user credentials inside support tools including links to the program certificate links, currently only available in the Django Admin of the Credentials app.

Link to the ticket: [PROD-2569](https://openedx.atlassian.net/browse/PROD-2569) 

Testing methodology:

- Go to the credentials app and create a [user credential](http://localhost:18150/admin/credentials/usercredential/ )
- Now you can view the newly created credentials inside the learner's tab within [support tools](http://localhost:18450/learner_information)

Screenshots:
Screenshot 2022-05-30 at 10 29 35 AM
Screenshot 2022-05-30 at 10 29 42 AM

### DIFF
--- a/src/users/LearnerCredentials.jsx
+++ b/src/users/LearnerCredentials.jsx
@@ -1,0 +1,133 @@
+import PropTypes from 'prop-types';
+import {
+  Hyperlink, TransitionReplace, Button, Alert,
+} from '@edx/paragon';
+import React, { useEffect, useState } from 'react';
+import Table from '../components/Table';
+import { getUserProgramCredentials } from './data/api';
+
+export default function LearnerCredentials({ username }) {
+  const [credentials, setCredentials] = useState(null);
+  const [error, setError] = useState(null);
+  const [isOpen, setIsOpen] = useState(false);
+
+  const Attributes = ({ attributes }) => (
+    <>
+      <TransitionReplace>
+        {isOpen ? (
+          <>
+            <Button
+              variant="link"
+              size="inline"
+              onClick={() => setIsOpen(false)}
+            >
+              Hide
+            </Button>
+            <Table
+              columns={[
+                {
+                  Header: 'Name',
+                  accessor: 'name',
+                },
+                {
+                  Header: 'Value',
+                  accessor: 'value',
+                },
+              ]}
+              data={attributes.map((attribute) => ({
+                name: attribute.name,
+                value: attribute.value,
+              }))}
+              styleName="custom-table"
+            />
+          </>
+        ) : (
+          <Button variant="link" size="inline" onClick={() => setIsOpen(true)}>
+            Show
+          </Button>
+        )}
+      </TransitionReplace>
+    </>
+  );
+
+  Attributes.propTypes = {
+    attributes: PropTypes.arrayOf(
+      PropTypes.shape({
+        name: PropTypes.string.isRequired,
+        value: PropTypes.string.isRequired,
+      }),
+    ).isRequired,
+  };
+
+  useEffect(() => {
+    if (username) {
+      getUserProgramCredentials(username).then((response) => {
+        if (response.count > 0) {
+          setCredentials(response.results);
+        } else if (response.errors) {
+          setError(response.errors[0]);
+        }
+      });
+    }
+  }, [username]);
+
+  return (
+    <>
+      <section className="mb-3">
+        <h3>Learner Credentials</h3>
+        {credentials ? (
+          <Table
+            columns={[
+              {
+                Header: 'Credential Type',
+                accessor: 'credentialType',
+              },
+              {
+                Header: 'Program ID',
+                accessor: 'programID',
+              },
+              {
+                Header: 'Status',
+                accessor: 'status',
+              },
+              {
+                Header: 'Certificate Link',
+                accessor: 'certificateUrl',
+              },
+              {
+                Header: 'Attributes',
+                accessor: 'attributes',
+              },
+            ]}
+            data={credentials.map((credential) => ({
+              credentialType: credential.credential.type,
+              programID: credential.credential.program_uuid,
+              status: credential.status,
+              certificateUrl: (
+                <Hyperlink destination={credential.certificate_url}>
+                  {credential.uuid}
+                </Hyperlink>
+              ),
+              attributes: <Attributes attributes={credential.attributes} />,
+            }))}
+            styleName="custom-table"
+          />
+        ) : (
+          <>
+            {error ? (
+              <>
+                <Alert variant="danger">{error.text}</Alert>
+              </>
+            ) : (
+              <p className="ml-4">No Credentials were Found.</p>
+            )}
+          </>
+        )}
+      </section>
+    </>
+  );
+}
+
+LearnerCredentials.propTypes = {
+  username: PropTypes.string.isRequired,
+};
--- a/src/users/data/api.js
+++ b/src/users/data/api.js
@@ -641,3 +641,29 @@ export async function regenerateCertificate(username, courseKey) {
     };
   }
 }
+
+export async function getUserProgramCredentials(username, page = 1) {
+  try {
+    const { data } = await getAuthenticatedHttpClient().get(
+      `${AppUrls.getUserCredentialsUrl()}?username=${username}&type=program&page=${page}`,
+    );
+    if (data.next !== null) {
+      const nextPageData = await getUserProgramCredentials(username, page + 1);
+      data.results = data.results.concat(nextPageData.results);
+      return data;
+    }
+    return data;
+  } catch (error) {
+    return {
+      errors: [
+        {
+          code: null,
+          dismissible: true,
+          text: error.text ? error.text : 'There was an error retrieving credentials for the user',
+          type: 'danger',
+          topic: 'credentials',
+        },
+      ],
+    };
+  }
+}
--- a/src/users/data/api.test.js
+++ b/src/users/data/api.test.js
@@ -6,6 +6,7 @@ import { enrollmentsData } from './test/enrollments';
 import { downloadableCertificate } from './test/certificates';
 import verifiedNameHistoryData from './test/verifiedNameHistory';
 import OnboardingStatusData from './test/onboardingStatus';
+import { credentials } from './test/credentials';
 import * as api from './api';
 import * as urls from './urls';
 import * as messages from '../../userMessages/messages';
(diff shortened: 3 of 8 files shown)


src/users/data/api.js
```diff
@@ -641,3 +641,24 @@ export async function regenerateCertificate(username, courseKey) {
...
+      `${AppUrls.getUserCredentialsUrl()}?username=${username}`,
+    );
+    return data;
+  } catch (error) {
+    return {
+      errors: [
+        {
+          code: null,
+          dismissible: true,
+          text: 'There was an error in retrieving credentials for the user',
```
DawoudSheraz: nit: if we can use error.text or a similar attribute instead of a hard-coded error message, that will help the user know the real issue on UI instead of going to the network tab.

src/users/data/api.js
```diff
@@ -641,3 +641,24 @@ export async function regenerateCertificate(username, courseKey) {
     };
   }
 }
+
+export async function getUserCredentials(username) {
+  try {
+    const { data } = await getAuthenticatedHttpClient().get(
+      `${AppUrls.getUserCredentialsUrl()}?username=${username}`,
+    );
+    return data;
```
DawoudSheraz: since the API is paginated, are we not catering if more than one page is returned in the response?

src/users/LearnerCredentials.jsx
```diff
@@ -0,0 +1,122 @@
...
+    }
+  }, [username]);
+
+  return (
+    <>
+      <section className="mb-3">
+        <h3>Learner Credentials</h3>
+        {credentials ? (
+          <Table
+            columns={[
```
DawoudSheraz, line 80: Where are the API call related errors rendered?

DawoudSheraz (commented):
added some comments/questions.

src/users/LearnerCredentials.jsx
```diff
@@ -0,0 +1,122 @@
...
+                  {credential.uuid}
+                </Hyperlink>
+              ),
+              attributes: <Attributes attributes={credential.attributes} />,
+            }))}
+            styleName="custom-table"
+          />
+        ) : (
+          <>
+            <p className="ml-4">No Credentials were Found.</p>
```
hurtstotouchfire: @oliviaruizknott I'm assuming all our bugs will just hit this line, but I'm wondering what happens in our certificate availability date cases. Possibly we could add more data here to help support identify that issue.

hurtstotouchfire:
Hey folks, I think any code review from Aperture here (Olivia and myself) can be considered non-blocking, and potentially we will follow up with our own PR. Thanks for building this! 🙏

src/users/LearnerCredentials.jsx
oliviaruizknott: Yes, in the future, I think we will want to add more data here; but this can be okay for now.

src/users/LearnerCredentials.jsx
```diff
@@ -0,0 +1,122 @@
...
+              },
+            ]}
+            data={credentials.map((credential) => ({
+              credentialType: credential.credential.type,
+              programID: credential.credential.program_uuid,
+              status: credential.status,
+              certificateUrl: (
+                <Hyperlink destination={credential.certificate_url}>
+                  {credential.uuid}
+                </Hyperlink>
```
oliviaruizknott, line 109: If you don’t filter by `type = 'program'`, this will render lots of broken / bad links.

The Credentials service only _renders_ **program** credentials. Course certificates are rendered by edx-platform.

In other words, even though all these credential records will have a `credential.certificate_url`, only credentials where `type = "program"` will have valid / working urls. Where `type = "course-run"`, the `certificate_url` will 404.

src/users/LearnerCredentials.jsx
```diff
@@ -0,0 +1,122 @@
...
+                accessor: 'certificateUrl',
+              },
+              {
+                Header: 'Attributes',
+                accessor: 'attributes',
+              },
+            ]}
+            data={credentials.map((credential) => ({
+              credentialType: credential.credential.type,
+              programID: credential.credential.program_uuid,
```
oliviaruizknott, line 104: Only credentials with `type = 'program'` will have this.

src/users/LearnerCredentials.jsx
```diff
@@ -0,0 +1,122 @@
...
+            />
+          </>
+        ) : (
+          <Button variant="link" size="inline" onClick={() => setIsOpen(true)}>
+            Show
+          </Button>
+        )}
+      </TransitionReplace>
+    </>
+  );
```
oliviaruizknott, line 51: nit [non-blocking]: For what it’s worth, this might be overkill; the only attributes we have stored are:
```
                {
                    "name": "visible_date",
                    "value": < some date value >
                }
```

There are no other Attribute records other than `visible_date` in the whole database. You _could_ simplify this so that the table just has a `visible_date` column, rather than handling for any attribute.

But this works too!

oliviaruizknott (changes requested):
It is possible that I read too hastily and you are doing this filtering somewhere that I missed. In that case, forgive me! But:

It looks like your API call is retrieving _all_ `UserCredential`s for the user from the Credentials service.

The Credentials service keeps `UserCredential` records for **both course and program** credentials. That means you will get back mostly **course** credentials in your call (simply because there are more of them), which doesn’t seem to be the intent.

If the intent is to quickly show / link to **program** credentials / certificates, filter your call or the results by `type = 'program'`.

Also: I didn’t have time to get set up to pull this down and click-test around locally myself. You could test this by creating a UserCredential with a `Credential content type` of `credentials | Course certificate configuration`. (Instead of `credentials | Program certificate configuration`.)

Screen Shot 2022-04-27 at 3 41 41 PM

How does that credential render in the UI? Do the links work? Does it behave as you expect / want it to?

While you’re there, **would you mind adding a screenshot to this PR showing this new UI in the Support Tool?**

DawoudSheraz:
Thank you for a detailed review and context @oliviaruizknott and @hurtstotouchfire. We will address the items that have been raised.

ansabgillani:
Apologies for the delayed response @oliviaruizknott and @hurtstotouchfire. Thank you for the detailed review. I will add the relevant checks and get back to you asap.

src/users/data/api.js
DawoudSheraz: @ansabgillani Still unresolved.

src/users/data/api.test.js
```diff
@@ -984,4 +986,28 @@ describe('API', () => {
...
+      type: 'danger',
+      topic: 'credentials',
+    };
+
+    it('Successful Credentials fetch', async () => {
+      mockAdapter.onGet(`${credentialUrl}?username=${testUsername}&type=program&page=1`).reply(200, successResponse);
+      const response = await api.getUserProgramCredentials(testUsername);
+      expect(response.count).toEqual(successResponse.count);
+      const { results } = response;
+      expect(results[0]).toEqual(successResponse.results[0]);
```
DawoudSheraz, line 1005: verify the paginated results concat in a separate unit test.

src/users/LearnerCredentials.jsx
```diff
@@ -0,0 +1,122 @@
...
+        value: PropTypes.string.isRequired,
+      }),
+    ).isRequired,
+  };
+
+  useEffect(() => {
+    if (username) {
+      getUserProgramCredentials(username).then((response) => {
+        if (response.count > 0) {
+          setCredentials(response.results);
```
DawoudSheraz, line 66: What happens if count is 0?

DawoudSheraz (commented):
- to add a screenshot of credentials view in the PR description
- Small nits to address before final approval

src/users/LearnerCredentials.jsx
ansabgillani: Resolved it out

src/users/LearnerCredentials.jsx
ansabgillani: Resolved it out

src/users/data/api.test.js
ansabgillani: added.

src/users/LearnerCredentials.jsx
ansabgillani: This means no credentials were present. So, the screen would display a 'No Credentials were found' error message

src/users/LearnerCredentials.jsx
ansabgillani: Not addressing these errors for now because of this https://github.com/openedx/frontend-app-support-tools/pull/253/files#r860233589

src/users/LearnerCredentials.jsx
DawoudSheraz: What happens if there is an error when fetching credentials? Where will the user see them?

DawoudSheraz: The above API call is only handling the cases when there is no error in the API call.

DawoudSheraz (commented):
1-2 questions added.

src/users/data/api.js
```diff
@@ -641,3 +641,29 @@ export async function regenerateCertificate(username, courseKey) {
...
+      return data;
+    }
+    return data;
+  } catch (error) {
+    return {
+      errors: [
+        {
+          code: null,
+          dismissible: true,
+          text: error.text ? error.text : 'There was an error in retrieving credentials for the user',
```
Ali-D-Akbar: nit: error retrieving

Ali-D-Akbar (commented):
Honestly, I think we should use the sub-table approach we used in enrollment/entitlement. The current approach really messes up the credential table.

Screenshot 2022-05-31 at 2 10 57 AM

src/users/LearnerCredentials.jsx
```diff
@@ -0,0 +1,122 @@
...
+    <>
+      <TransitionReplace>
+        {isOpen ? (
+          <>
+            <Button
+              variant="link"
+              size="inline"
+              onClick={() => setIsOpen(false)}
+            >
+              Hide
```
justinhynes, line 24: Nit: Since this is in the `openedx` org, which makes me believe this is used outside of edX/2U, does the Support Tools MFE support localization? Should `Hide` be hard-coded English?

src/users/LearnerCredentials.jsx
```diff
@@ -0,0 +1,122 @@
...
+              data={attributes.map((attribute) => ({
+                name: attribute.name,
+                value: attribute.value,
+              }))}
+              styleName="custom-table"
+            />
+          </>
+        ) : (
+          <Button variant="link" size="inline" onClick={() => setIsOpen(true)}>
+            Show
```
justinhynes, line 46: Nit: See above comment on localization of `Hide`. Should `Show` be hard-coded English?

src/users/LearnerCredentials.jsx
```diff
@@ -0,0 +1,122 @@
...
+                  {credential.uuid}
+                </Hyperlink>
+              ),
+              attributes: <Attributes attributes={credential.attributes} />,
+            }))}
+            styleName="custom-table"
+          />
+        ) : (
+          <>
+            <p className="ml-4">No Credentials were Found.</p>
```
justinhynes: Same question here (see above comment); Should `No Credentials were Found` be hard-coded English?

src/users/LearnerCredentials.jsx
```diff
@@ -0,0 +1,122 @@
...
+          setCredentials(response.results);
+        }
+      });
+    }
+  }, [username]);
+
+  return (
+    <>
+      <section className="mb-3">
+        <h3>Learner Credentials</h3>
```
justinhynes, line 77: Nit: Same question here, should `Learner Credentials` be hard-coded English?

oliviaruizknott (approved):
I re-reviewed for the handling of _program_ certificates only and am satisfied by the filtering in the API call! Thanks.

(Though I’m trusting that you checked that this actually works; I haven’t checked the Credentials side.)

DawoudSheraz (approved):
Looks good codewise. One improvement for a follow-up PR:
- Use table component expandedrow behavior to display additional attributes

PR #257, opened 2022-06-21
fix(header): update urls

azanbinzahid:
### Description:
This PR updates outdated urls in header

### Linked Ticket:
[PROD-2868](https://2u-internal.atlassian.net/browse/PROD-2868)

### Merge After:
https://github.com/edx/edx-internal/pull/6769

### Testing:
Check if header buttons go to correct URL as requested in ticket.


src/supportHeader/Header.jsx
```diff
@@ -45,22 +45,17 @@ export default function Header() {
...
+      href: 'https://2u-internal.atlassian.net/wiki/spaces/SPRT/overview',
     },
     {
       type: 'submenu',
       content: 'Escalations',
       submenuContent: (
         <>
           <div className="mb-1"><a rel="noopener" href="https://edx.lightning.force.com/lightning/o/Case/list?filterName=00B0L000005D8BgUAK">Salesforce</a></div>
-          <div className="mb-1"><a rel="noopener" href="https://openedx.atlassian.net/servicedesk/customer/user/requests?page=1&status=open">CR Requests</a></div>
+          <div className="mb-1"><a rel="noopener" href="https://2u-internal.atlassian.net/servicedesk/customer/portal/9">CR Requests</a></div>
```
DawoudSheraz: If we can have environment vars for these that we define in edx-internal and use them here. These 2U-specific urls in openedx repo might not be a good idea.

azanbinzahid: We can do that, just need to add them [here](https://github.com/edx/edx-internal/tree/master/frontends/frontend-app-support-tools).

DawoudSheraz (approved):

PR #268, opened 2022-07-26
fix: pdf cert url behavior

DawoudSheraz:
### Description
The certificate component is appending LMS base url against download_url field of certificate data. However, the field download_url contains a complete url if the field is_pdf_certificate is true. This caused PDF download url to form incorrectly on support tools. This PR is fixing that behavior. Reference: https://github.com/openedx/edx-platform/blob/master/lms/djangoapps/certificates/api.py#L61

### DIFF
--- a/src/users/data/test/certificates.js
+++ b/src/users/data/test/certificates.js
@@ -17,3 +17,14 @@ export const regeneratableCertificate = {
   downloadUrl: null,
   regenerate: true,
 };
+
+export const pdfCertificate = {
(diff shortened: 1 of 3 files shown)


azanbinzahid (approved):
Do we need to add method description for `certificateDownloadUrl` explaining the reason?

Ali-D-Akbar (approved):

DawoudSheraz:
> Do we need to add method description for `certificateDownloadUrl` explaining the reason?

I can add that.

PR #269, opened 2022-08-03
build: remove check-prop-types from support tools to fix dependency issues

DawoudSheraz:
### Description
- Remove check-prop-types and its usage in support tools to fix its dependency conflict with prop-types in frontend-platform
- This package is not actively maintained. Plus, its usage is very limited in the repository. It is safe to remove this
- Apart from this, a small update in ProgramInspector tests to avoid log statements after test suite execution.

### Testing
- Checkout master branch of support tools. Run npm install twice. You will see failure due to prop-type version conflict
- Check out this PR and do npm install twice. Both runs will be successful with no error reported.


ansabgillani (approved):

azanbinzahid (approved):
I see this was basically a test dependency.

PR #273, opened 2022-09-28
feat: learner record tab

MaxFrank13:
This PR adds a new "Learner Records" tab under the Learner Information tab. This new tab contains information about a Learner's program records. 

Screen Shot 2022-09-28 at 12 52 51 PM

This feature is intended to replace masquerading for the [current Learner Records page](https://credentials.edx.org/records/).  Once we convert this page to an MFE, it will no longer support masquerading and this tab will be provided as a way for the support team to look up a Learner's records.

Screen Shot 2022-09-28 at 12 54 33 PM

NOTE: This feature _does not_ allow for the creation of public links. The option to copy a link will only appear if a public record exists for that learner (captured in the `shared_program_record_uuid` field of the initial API response). 

There is a `USE_LEARNER_RECORD_TAB` variable that acts as a configuration flag to toggle the tab. As well as a `RECORDS_BASE_URL` variable that is used by the copy button when building a public link.


.env.development
```diff
@@ -19,8 +19,10 @@ LOGO_WHITE_URL=https://edx-cdn.org/v3/default/logo-white.svg/
 FAVICON_URL=https://edx-cdn.org/v3/default/favicon.ico
 MARKETING_SITE_BASE_URL='http://localhost:18000'
 ORDER_HISTORY_URL='localhost:1996/orders'
+RECORDS_BASE_URL='localhost:1990'
```
justinhynes, line 22: Do we need defaults for these settings in the `.env` file as well?

src/users/LearnerRecords.jsx
```diff
@@ -0,0 +1,126 @@
...
+    }
+  }, [username]);
+
+  const handleCopyButton = (uuid) => {
+    navigator.clipboard.writeText(`${RECORDS_BASE_URL}/shared/${uuid}`);
+  };
+
+  return (
+    <section>
+      <h3>Learner Records</h3>
```
justinhynes: I was under the impression that we should always be starting from an `H1` and moving down. I see that this is contained within a "`section`" component, so maybe there is already an H1 in the hierachy? Should this be "styled like" an H3 using CSS rather than using the `H3` tag?

Maybe a11y isn't as important as this is an internal support tool?

(Honestly just asking because I am trying to flex some "rules" I _think_ I understand, but it's an honest question for my own edification)

src/users/LearnerRecords.jsx
```diff
@@ -0,0 +1,126 @@
...
+      {
+        records ? (
+          records.map(({ record, uuid }, idx) => (
+            <section key={uuid} className={`${idx % 2 ? 'bg-light-100' : 'bg-light-200'} p-4`}>
+              <div className="d-flex align-items-center justify-content-between">
+                <div>
+                  <h4>{record.program.name}</h4>
+                  <p>{record.program.type_name}</p>
+                  <p>{record.program.school} | {record.platform_name}</p>
```
justinhynes: Do we _need_ the `school` and `platform_name` here? It makes sense on the Program Record page, but I'm not sure it's useful here. I could see having the program type and program name.

src/users/LearnerRecords.jsx
```diff
@@ -0,0 +1,126 @@
...
+                  <p>Last updated: {new Date(record.program.last_updated).toLocaleDateString()}</p>
+                </div>
+                {record.shared_program_record_uuid && (
+                  <Button
+                    variant="primary"
+                    onClick={() => handleCopyButton(record.shared_program_record_uuid.replaceAll('-', ''))}
+                  >
+                    Copy public record link
```
justinhynes: Should this be hard-coded english? This tool exists as part of the `openedx` GitHub org and it _could_ be used by other Open edX operators. 

My assumption here is that we'd want to resource the string.

src/users/LearnerRecords.jsx
```diff
@@ -0,0 +1,126 @@
...
+      {
+        records ? (
+          records.map(({ record, uuid }, idx) => (
+            <section key={uuid} className={`${idx % 2 ? 'bg-light-100' : 'bg-light-200'} p-4`}>
+              <div className="d-flex align-items-center justify-content-between">
+                <div>
+                  <h4>{record.program.name}</h4>
+                  <p>{record.program.type_name}</p>
+                  <p>{record.program.school} | {record.platform_name}</p>
+                  <p>Last updated: {new Date(record.program.last_updated).toLocaleDateString()}</p>
```
justinhynes: See below comment on having hard-coded english here.

src/users/LearnerRecords.jsx
```diff
@@ -0,0 +1,126 @@
...
+                      ? `${parseInt(Math.round(grade.percent_grade * 100), 10).toString()}%`
+                      : '',
+                  issue_date:
+                    grade.issue_date
+                      ? new Date(grade.issue_date).toLocaleDateString()
+                      : '',
+                  status:
+                    grade.issue_date
+                      ? 'Earned'
+                      : 'Not Earned',
```
justinhynes: See above comment on hard-coded english.

src/users/LearnerRecords.jsx
```diff
@@ -0,0 +1,126 @@
...
+                    onClick={() => handleCopyButton(record.shared_program_record_uuid.replaceAll('-', ''))}
+                  >
+                    Copy public record link
+                  </Button>
+                )}
+              </div>
+              <Table
+                columns={[
+                  {
+                    Header: 'Course Name',
```
justinhynes: For each `Header` in this section from lines 55 - 83, should these be resourced as well?

src/users/LearnerRecords.jsx
```diff
@@ -0,0 +1,126 @@
...
+                styleName="custom-table"
+              />
+            </section>
+          ))
+        ) : (
+          <>
+            {error ? (
+              <Alert variant="danger">{error.text}</Alert>
+            ) : (
+              <p>No results found for username: {username}</p>
```
justinhynes: Same hardcoded-english observation/question here.

src/users/LearnerRecords.jsx
MaxFrank13: I had thought so as well but when I initially looked over the other components I saw all hard-coded english. It looks like they are resourcing in some parts with the `FormattedMessage` component. Given that the component is available, it should be easy enough to modify at least the strings I've added here. I'll get on those changes 👍

src/users/data/test/records.js
```diff
@@ -0,0 +1,56 @@
+const records = [
```
Tj-Tracy, line 1: Not worthwhile to change now, but in the future, we tend to use Rosie for this kind of test data: https://github.com/rosiejs/rosie

Just something to keep in mind for the next time you're doing something like this

Tj-Tracy (approved):
LGTM 👍 One comment around test data, but not blocking.

justinhynes:
@MaxFrank13  -- Are we missing the status of the Program Certificate in the table?

src/users/LearnerRecords.jsx
justinhynes: Thank you!!

justinhynes (changes requested):
I think we'd want to display the overall status of the Program certificate as well in the tables.

src/users/LearnerRecords.jsx
MaxFrank13: It's my understanding that `h1` is used for the most important title on the _entire_ page. I don't believe there actually is one for this page. The "Support Tools" string at the top is an `h2`. All of the other headers for the tabs were `h3` so that's why I settled on it. I think it makes sense as an `h3` given that is a subsection of the page. However, I'm not well-versed on a11y

src/users/LearnerRecords.jsx
MaxFrank13: Going to replace this line with the "Status" of the record

src/users/LearnerRecords.jsx
justinhynes: If this were a learner/partner facing page, I'd be inclined to chase down this answer. For now, I think this is sufficient. Thanks, Max.

MaxFrank13:
> I think we'd want to display the overall status of the Program certificate as well in the tables.

Added 👍 

Screen Shot 2022-09-28 at 3 33 17 PM

justinhynes (approved):
Thanks for all the quick changes! 👍

src/users/LearnerRecords.jsx
```diff
@@ -0,0 +1,152 @@
...
+        intl.formatMessage(messages.earnedStatus)
+      );
+    }
+    if (program.empty) {
+      return (
+        intl.formatMessage(messages.notEarnedStatus)
+      );
+    }
+    return (
+      intl.formatMessage(messages.partiallyCompletedStatus)
```
DawoudSheraz, line 44: nitt: you can use JSON here, with status as keys and expected message as value to remove if conditions.

src/users/data/test/records.js
DawoudSheraz: Oh, this is great. Factory-boy for JS tests it seems.

src/users/data/api.js
```diff
@@ -698,3 +698,29 @@ export async function getUserProgramCredentials(username, page = 1) {
...
+export async function getLearnerRecords(username) {
+  try {
+    const { data } = await getAuthenticatedHttpClient().get(`${AppUrls.getLearnerRecordsUrl()}/?username=${username}`);
+    const programDetails = [];
+
+    if (data.enrolled_programs.length > 0) {
+      await Promise.all(data.enrolled_programs.map(program => getAuthenticatedHttpClient().get(`${AppUrls.getLearnerRecordsUrl()}/${program.uuid}/?username=${username}`)
+        .then(response => programDetails.push(response.data))));
+    }
+
```
DawoudSheraz: What should be the behavior here if one of the program record calls is not successful and there is an error message returned?

src/users/LearnerRecords.test.jsx
```diff
@@ -0,0 +1,144 @@
...
+    };
+    apiMock = jest
+      .spyOn(api, 'getLearnerRecords')
+      .mockImplementationOnce(() => Promise.resolve(expectedError));
+    wrapper = mount(<LearnerRecordsWrapper username={data.username} />);
+
+    await waitForComponentToPaint(wrapper);
+
+    expect(wrapper.find('.alert').text()).toEqual(expectedError.errors[0].text);
+    wrapper.unmount();
```
DawoudSheraz: nit: the unmount can be moved to afterEach.

DawoudSheraz (commented):
Overall, the changes are good. Can you add testing steps, including some pointers on setting test data in credentials, so that I can test it out on local, please? Thank you

src/users/data/api.js
MaxFrank13: This error would be caught by the `catch` below it. That error message is then displayed as an alert in the UI. I will add a test that demonstrates what happens when one of the calls in the `Promise.all()` fails 👍

MaxFrank13:
> Overall, the changes are good. Can you add testing steps, including some pointers on setting test data in credentials, so that I can test it out on local, please? Thank you

In order to test the new tab UI, you'll need:

- A user who is enrolled in at least one course in a program and earned a certificate for that program. This is necessary for the REST API to return any data. I used myself (username=edx) when testing this locally.

  - Make sure you have pulled the latest version of Credentials to have access to this REST API

- Add new environment variables. Create and set a `USE_LEARNER_RECORD_TAB` variable to `true` and then `RECORDS_BASE_URL` to `localhost:1990`

From there, you should be able to see a list of the user's earned records under the Learner Records tab. 

- If a public version of the record exists, there will be a button for copying a link to that record. Click the "Copy public record link" and make sure a link copies to your clipboard. Note that in order for this link to work, you'll need to have the latest version Learner Record MFE running.

  - If you want to create a public record, go to the Learner Record page (Learner Record MFE), click on "View Program Record" for an enrolled program, and  select "Create public record link"
  - Refreshing the support tool should render a button for copying the link
  
-  The REST API can also be tested to verify that the data is coming through at: 

   - `http://localhost:18150/records/api/v1/program_records/{program_id}/?username={username}`
   - NOTE: the REST API has been deployed to Production

src/users/LearnerRecords.jsx
MaxFrank13: Could you elaborate further on this? I think I understand but am having trouble piecing it together. The `program.completed` and `program.empty` fields are both booleans. My thinking is to convert these into JSON keys who's values correspond to their appropriate message, but how are you suggesting to do this?

src/users/LearnerRecords.test.jsx
ansabgillani: Refactor still needs to be addressed.

src/users/data/api.js
```diff
@@ -698,3 +698,29 @@ export async function getUserProgramCredentials(username, page = 1) {
...
   }
 }
+
+export async function getLearnerRecords(username) {
+  try {
+    const { data } = await getAuthenticatedHttpClient().get(`${AppUrls.getLearnerRecordsUrl()}/?username=${username}`);
+    const programDetails = [];
+
+    if (data.enrolled_programs.length > 0) {
+      await Promise.all(data.enrolled_programs.map(program => getAuthenticatedHttpClient().get(`${AppUrls.getLearnerRecordsUrl()}/${program.uuid}/?username=${username}`)
```
ansabgillani: nit: Can we maybe try to make it indented for better readability?

ansabgillani:
The changes look good to me. A couple of refactors that may improve the code readability, otherwise good to go.

ansabgillani (approved):

src/users/LearnerRecords.jsx
DawoudSheraz: Oh, ok. I mistook them for strings.

PR #275, opened 2022-10-18
feat: alert for private record

MaxFrank13:
This will add behavior that renders an alert when a record does not have a public link. This is in place of the "Copy public record link" button that would render if a public link exists for that record.

public_record_alert

Screen Shot 2022-10-18 at 1 10 36 PM


justinhynes (approved):
LGTM! Seems straight forward.

DawoudSheraz (approved):

Ali-D-Akbar (approved):

Ali-D-Akbar (approved):

PR #276, opened 2022-10-18
fix: don't generate frontend specific urls for the Learner Record MFE

justinhynes:
[APER-1975]

All links to public program records will point to the Credentials IDA and the backend will figure out whether it should be redirected to the legacy FE or the LR MFE. This also removes the `RECORDS_BASE_URL` setting which we no longer need.


MaxFrank13 (approved):
Looks good. I just merged my changes to this MFE but I don't think there are any conflicts (I assume Github would scream at us?)

justinhynes:
Yeah, you're right. If there were any conflicts, it _should_ get picked up here. :)

PR #278, opened 2022-10-24
chore: removing Learner Record feature flag

MaxFrank13:
This will remove the feature flag for the Learner Records tab. The feature flag was set up incorrectly and was not working as intended resulting in the string "null" being read as a truth-y value by JavaScript. 

The feature flag was originally intended to safeguard against providing links to an MFE that wasn't enabled (the Learner Record MFE). However, this was rectified in [this PR](https://github.com/openedx/frontend-app-support-tools/pull/276) where the generated link was modified to point to Credentials. The decision to redirect is then handled by [Credentials in the backend](https://github.com/openedx/credentials/pull/1793).


cdeery (approved):
Looks good!

Tj-Tracy (approved):

PR #314, opened 2023-03-29
feat: add support for initiating retirement flow

zawan-ila:
# [PROD-3201](https://2u-internal.atlassian.net/browse/PROD-3201)

## Description

This PR adds support for retiring users.

## Pictures
Screenshot 2023-03-29 at 7 55 29 AM

## Demo

https://user-images.githubusercontent.com/87228907/229508270-23226edd-9525-4c00-8f04-726d8b4ea3fe.mov

## New Positioning of Button

Screenshot 2023-04-12 at 11 12 30 AM

### DIFF
--- a/src/users/account-actions/AccountActions.jsx
+++ b/src/users/account-actions/AccountActions.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 
 import TogglePasswordStatus from './TogglePasswordStatus';
 import ResetPassword from './ResetPassword';
+import RetireUser from './RetireUser';
 import PasswordHistory from './PasswordHistory';
 
 export default function AccountActions({ userData, changeHandler }) {
@@ -15,6 +16,11 @@ export default function AccountActions({ userData, changeHandler }) {
         changeHandler={changeHandler}
       />
       <ResetPassword
+        email={userData.email}
+        changeHandler={changeHandler}
+      />
+      <RetireUser
+        username={userData.username}
         email={userData.email}
         changeHandler={changeHandler}
       />
--- a/src/users/account-actions/RetireUser.jsx
+++ b/src/users/account-actions/RetireUser.jsx
@@ -1,0 +1,112 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import { FormattedMessage } from '@edx/frontend-platform/i18n';
+import {
+  Button, Alert, ModalDialog, ActionRow,
+} from '@edx/paragon';
+import { postRetireUser } from '../data/api';
+
+export default function RetireUser({
+  username,
+  email,
+  changeHandler,
+}) {
+  const [retireUserModalIsOpen, setRetireUserModalIsOpen] = useState(false);
+  const [errorMessage, setErrorMessage] = useState(null);
+
+  const isNotRetirable = () => {
+    if (email.endsWith('@retired.invalid')) { return true; }
+    return false;
+  };
+  const retireUser = async () => {
+    const resp = await postRetireUser(username);
+    if (resp.errors) {
+      setErrorMessage(resp.errors[0].text || 'Something went wrong. Please try again later!');
+    } else {
+      changeHandler();
+    }
+  };
+
+  const closeRetireUserModal = () => {
+    setRetireUserModalIsOpen(false);
+    setErrorMessage(null);
+  };
+
+  const modalBody = (
+    errorMessage ? <Alert variant="danger">{errorMessage}</Alert>
+      : (
+        <div>
+          <Alert variant="warning">
+            <FormattedMessage
+              id="supportTools.accountActions.retireUser"
+              tagName="p"
+              description="User retirement confirmation prompt"
+              defaultMessage="You are about to retire {username} with the email address: {email}.{br}{br}
+              This is a serious action that will revoke this user's access to edX and
+              their earned certificates. Furthermore, the email address associated
+              with the retired account will not be able to be used to create a new account."
+              values={{
+                username: <strong>{username}</strong>,
+                email: <strong>{email}</strong>,
+                br: <br />,
+              }}
+            />
+          </Alert>
+        </div>
+      )
+  );
+
+  return (
+    <div>
+      <Button
+        id="retire-user"
+        variant="btn btn-danger"
+        onClick={() => setRetireUserModalIsOpen(true)}
+        className="mr-1 mb-2 ml-2"
+        disabled={isNotRetirable()}
+      >
+        Retire User
+      </Button>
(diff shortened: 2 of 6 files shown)


src/users/data/api.js
```diff
@@ -98,14 +98,16 @@ export async function getUser(userIdentifier) {
   const url = AppUrls.getUserAccountUrl(userIdentifier);
 
   try {
-    const { data } = await getAuthenticatedHttpClient().get(url);
+    const resp = await getAuthenticatedHttpClient().get(url);
+    console.log('getUser', resp);
```
DawoudSheraz: unnecessary log statement.

src/users/data/api.js
```diff
@@ -98,14 +98,16 @@ export async function getUser(userIdentifier) {
...
+    const { data } = resp;
     return Array.isArray(data) && data.length > 0 ? data[0] : data;
   } catch (error) {
     // We don't have good error handling in the app for any errors that may have come back
     // from the API, so we log them to the console and tell the user to go look.  We would
     // never do this in a customer-facing app.
     // eslint-disable-next-line no-console
-    console.log(JSON.parse(error.customAttributes.httpErrorResponseData));
+    console.log('eerrr', JSON.parse(error.customAttributes.httpErrorResponseData));
```
DawoudSheraz: 'eerrr' ?

src/users/data/api.js
```diff
@@ -575,6 +579,50 @@ export async function postResetPassword(email) {
...
+    const response = await getAuthenticatedHttpClient().post(
+      AppUrls.userRetirementUrl(),
+      {
+        usernames: username,
+      },
+    );
+
+    if (response.data.failed_user_retirements.length > 0) {
+      errMessage = 'Server Error. The backend service(lms) failed to retire the user';
+      throw new Error();
```
DawoudSheraz, line 590: Why is this throw needed?

src/users/data/api.js
zawan-ila: fixed

src/users/data/api.js
zawan-ila: fixed

src/users/data/api.js
zawan-ila: Because we want to catch this :-)

Sometimes the server fails to process a retirement request and indicates the failed retirements as a list(failed_user_retirements) in `response.data`. The status code is still 200. We wish to treat such conditions as errors and indicate so to the calling code.

This error is caught just a few lines below.

AliAdnanSohail (approved):

ProductRyan:
@zawan-ila three notes:

1. can you place the button so that its isolated and not easy to accidentally click? Perhaps to the right of the other 2 buttons in the box, with some additional padding.
2. New copy for the confirmation modal body

> You are about to retire [username] with the email address: [email] 
> 
> This is a serious action that will revoke this user's access to edX and their earned certificates. Furthermore, the email address associated with the retired account will not be able to be used to create a new account.

3. How long can a support user cancel the retirement? Perhaps we should add the timestamp that the retirement was requested on that page and tell the support user how long they have to cancel it.

zawan-ila (commented):
Hi @ProductRyan,
Regarding your comments above
1. I have moved the button to the right and added some padding too.
2. Modal body has been updated
3. Retirements can be cancelled within 14 days. This work was done some while ago and support is already aware of it. Also the cancel-retirement button does not show if the 14-day period has passed.

src/users/account-actions/RetireUser.jsx
```diff
@@ -0,0 +1,112 @@
...
+              tagName="p"
+              description="User retirement confirmation prompt"
+              defaultMessage="You are about to retire {username} with the email address: {email}.{br}{br}
+              This is a serious action that will revoke this user's access to edX and
+              their earned certificates. Furthermore, the email address associated
+              with the retired account will not be able to be used to create a new account."
+              values={{
+                username: <strong>{username}</strong>,
+                email: <strong>{email}</strong>,
+                br: <br />,
```
DawoudSheraz, line 51: why add br: <br/>?

src/users/account-actions/AccountActions.jsx
```diff
@@ -18,6 +19,11 @@ export default function AccountActions({ userData, changeHandler }) {
         email={userData.email}
         changeHandler={changeHandler}
       />
+      <RetireUser
+        username={userData.username}
+        email={userData.email}
+        changeHandler={changeHandler}
```
DawoudSheraz, line 25: what would be changeHandler here? Because once the retirement request is submitted, we can't get user information again, right? We only get option to cancel retirement. Please correct me if I am missing something.

src/users/account-actions/RetireUser.test.jsx
```diff
@@ -0,0 +1,98 @@
...
+    expect(confirmAlert.text()).toEqual(
+      "You are about to retire edx with the email address: edx@example.com. This is a serious action that will revoke this user's access to edX and their earned certificates. Furthermore, the email address associated with the retired account will not be able to be used to create a new account.",
+    );
+    retireUserModal.find('button.btn-danger').hostNodes().simulate('click');
+    await waitForComponentToPaint(wrapper);
+    expect(UserSummaryData.changeHandler).toHaveBeenCalled();
+    retireUserModal.find('button.btn-link').simulate('click');
+    retireUserModal = wrapper.find('ModalDialog#user-account-retire');
+    expect(retireUserModal.prop('isOpen')).toEqual(false);
+
```
DawoudSheraz, line 59: We should also assert that after reitrement request, the new response does not render acct details but the msg that acct belongs to a retired user.

DawoudSheraz (approved):
Overall, looks good. A few qs/nits to address before merge.

src/users/account-actions/RetireUser.jsx
zawan-ila: This represents line breaks. Since we need these in the modal message we need to introduce this here. See Ryan's comments above

src/users/account-actions/AccountActions.jsx
zawan-ila: the changehandler here is used to reload userInfo after we send some request that may change user info (e.g retiring them)...

src/users/account-actions/RetireUser.test.jsx
zawan-ila: Umm that would be testing multiple components then...As things stand now there is a separate component for the message `Acct belongs to retired User` and cancel retirement button. So this sort of testing would probably be need to be done in a parent component of both of these. Also the decision of which component to render is made after `changeHandler` is called(some details above) and we are already asserting on it being called.

PR #327, opened 2023-04-14
feat: check for valid term date format

ghost:
commit c5691989b993fb1d780ed16a90bf735e71430d69
Author: Emily Aquin <eaquin@2u.com>
Date:   Fri Apr 14 13:50:42 2023 -0500

feat: check for valid term date format

In the Support Tools: New Learner Credit Plan, a validity check was added that
    the term start and end date is in the format "YYYY-MM-DD".

    ENT-6940

### DIFF
--- a/src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormTerm.jsx
+++ b/src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormTerm.jsx
@@ -3,6 +3,7 @@ import { Form } from '@edx/paragon';
 import PROVISIONING_PAGE_TEXT from '../data/constants';
 import useProvisioningContext from '../data/hooks';
 import { selectProvisioningContext } from '../data/utils';
+import { isValidDateString } from '../../../utils';
 
 const ProvisioningFormTerm = () => {
   const { TERM } = PROVISIONING_PAGE_TEXT.FORM;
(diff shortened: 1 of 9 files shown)


src/utils/index.js
```diff
@@ -96,3 +97,5 @@ export function extractParams(searchParams) {
       .map(queryParams => queryParams.split('=')),
   );
 }
+
+export const isValidDateString = (dateString) => Boolean(dateString && dateString.match(DATE_REGEX) != null);
```
marlonkeating: Since we're pulling in the moment library, would it make sense to leverage its parsing capabilities?
https://momentjs.com/docs/#/parsing/string-format/

ghost: Great idea!

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormTerm.jsx
```diff
@@ -12,10 +13,14 @@ const ProvisioningFormTerm = () => {
...
-      return setStartDate(eventTarget.value);
+    if (isValidDateString(eventTarget.value)) {
+      if (eventTarget.dataset.testid.includes('start')) {
+        return setStartDate(eventTarget.value);
+      }
+      return setEndDate(eventTarget.value);
     }
-    return setEndDate(eventTarget.value);
+    setStartDate('');
+    return setEndDate('');
```
pwnage101: Why set both to null just if one of them might fail validation?  E.g. if end date was correct, but start date failed validation, end date would get unset, right?

pwnage101 (approved):
LGTM!  Might still want an approval from somebody who does more frontend.

src/Configuration/Provisioning/ProvisioningForm/tests/ProvisioningFormTerm.test.jsx
```diff
@@ -62,5 +62,13 @@ describe('ProvisioningFormTerm', () => {
     expect(endDateInputValue).toEqual('2020-01-01');
 
     expect(screen.getByText(TERM.VALIDITY)).toBeTruthy();
+
+    fireEvent.change(endDateInput, { target: { value: 'foo' } });
+    const invalidEndDateInputValue = endDateInput.getAttribute('value');
+    expect(invalidEndDateInputValue).toEqual('');
```
pwnage101, line 68: maybe this should also check if `expect(startDateInput.getAttribute('value')).toEqual('2021-01-01')` to validate that it's unchanged?

pwnage101 (approved):

PR #328, opened 2023-04-14
feat: integrates creation of enterprise customer catalog

brobro10000:
Adds endpoint and integration of creating enterprise customer catalogs


src/Configuration/Provisioning/data/hooks.js
```diff
@@ -78,11 +79,24 @@ export default function useProvisioningContext() {
 
   const hydrateCatalogQueryData = useCallback(async () => {
     const { data } = await LmsApiService.fetchEnterpriseCatalogQueries();
+    const camelCasedData = camelCaseObject(data.results);
+    const predeterminedCatalogQueries = camelCasedData.filter(({ title }) => !title.search(/\[DO NOT ALTER\]\[LEARNER CREDIT\]/));
```
pwnage101: Nit: if we switch to using `title.indexOf()` we can be more DRY and pull out the prefix into a const (or even in `../../data/constants`???) and used everywhere.  E.g.:
```suggestion
    const learnerCreditCatalogPrefix = '[DO NOT ALTER][LEARNER CREDIT]'
    const predeterminedCatalogQueries = camelCasedData.filter(
      ({ title }) => title.indexOf(learnerCreditCatalogPrefix) != 0
    );
```

and below:

```python
      if (catalogQueryObject.title.split(learnerCreditCatalogPrefix + ' ')[1] === catalogQuery.catalogQueryTitle.split(' account')[0]) {
```

src/Configuration/Provisioning/data/hooks.js
```diff
@@ -78,11 +79,24 @@ export default function useProvisioningContext() {
...
+    const filteredCourses = camelCasedData.filter(({ title }) => title.search(/\[DO NOT ALTER\]\[LEARNER CREDIT\]/));
+    INITIAL_CATALOG_QUERIES.multipleQueries.forEach((catalogQuery, index) => {
+      const catalogQueryObject = predeterminedCatalogQueries.filter(
+        ({ title }) => !title.search(new RegExp(`.*?${catalogQuery.catalogQueryTitle.split(' account')[0]}`)),
+      )[0];
+      if (catalogQueryObject.title.split('[DO NOT ALTER][LEARNER CREDIT] ')[1] === catalogQuery.catalogQueryTitle.split(' account')[0]) {
```
pwnage101: I'm a bit confused by this.  If `catalogQueryObject` is selected from `predeterminedCatalogQueries`, and `predeterminedCatalogQueries` contains only catalogs without the `[DO NOT ALTER][LEARNER CREDIT] ` prefix, then isn't `catalogQueryObject.title.split('[DO NOT ALTER][LEARNER CREDIT] ')[1]` guaranteed to be undefined?

src/Configuration/Provisioning/data/hooks.js
```diff
@@ -78,11 +79,24 @@ export default function useProvisioningContext() {
 
   const hydrateCatalogQueryData = useCallback(async () => {
     const { data } = await LmsApiService.fetchEnterpriseCatalogQueries();
+    const camelCasedData = camelCaseObject(data.results);
+    const predeterminedCatalogQueries = camelCasedData.filter(({ title }) => !title.search(/\[DO NOT ALTER\]\[LEARNER CREDIT\]/));
+    const filteredCourses = camelCasedData.filter(({ title }) => title.search(/\[DO NOT ALTER\]\[LEARNER CREDIT\]/));
```
pwnage101: Is this variable name correct?  Does the resulting list contain courses?

src/Configuration/Provisioning/data/hooks.js
brobro10000: Implemented the first comment solution as it removes the need to use regex's 👍🏽 . Once we brought in the ids through the configuration, it was a solution that allowed it to work on local stage and prod environments removing the need to filter twice.

src/Configuration/Provisioning/data/hooks.js
brobro10000: Yep, that was a snafu. Corrected now 👍🏽

src/Configuration/Provisioning/data/hooks.js
brobro10000: No longer using this convoluted logic. 😅  It wasn't a good implementation which is why I sought a different solution. This comment is an example why it wasn't a great implementation technique in the first place.

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormSubmissionButton.jsx
```diff
@@ -1,40 +1,94 @@
...
+        console.log(ALERTS.API_ERROR_MESSAGES.ENTERPRISE_CUSTOMER_CATALOG[customAttributes.httpErrorStatus]);
+      }
+    }
   };
+  useEffect(() => {
+    if (submitButtonState === 'complete') {
+      resetFormData();
+      history.push(HOME);
+    }
+  });
```
adamstankiewicz: Is it intentional to not have a dependency array on this `useEffect`? 👀  As is, it'd get executed on every re-render regardless of whether `submitButtonState` changes.

src/Configuration/Provisioning/data/constants.js
```diff
@@ -13,6 +13,9 @@ const PROVISIONING_PAGE_TEXT = {
     SUB_TITLE: 'Plan Details',
     BUTTON: {
       submit: 'Create Learner Credit Plan',
```
adamstankiewicz: nit: fwiw, typically our button CTAs only have the first letter capitalized.

src/Configuration/Provisioning/data/constants.js
```diff
@@ -13,6 +13,9 @@ const PROVISIONING_PAGE_TEXT = {
...
+      pending: 'Creating...',
+      success: 'Plan successfully created',
+      error: 'Failed to create plan',
```
adamstankiewicz: [curious] What is the user's course of action if/when an error occurs? Are they able to re-try somehow? For example, in the new `StatefulEnroll` button for EMET in enterprise learner portal, the error text is "Try again" and clicking the button makes another attempt.

I guess my concern is that "Failed to create plan" alone doesn't quite give the affordance to the user that they can click the button again to re-try? Food for thought.

src/data/services/EnterpriseApiService.js
```diff
@@ -6,9 +6,20 @@ class LmsApiService {
...
+  static enterpriseAPIBaseUrl = `${LmsApiService.baseUrl}/enterprise/api/v1/`;
+
+  static enterpriseCatalogQueriesUrl = `${LmsApiService.enterpriseAPIBaseUrl}enterprise_catalog_query/`;
+
+  static enterpriseCustomerCatalogsUrl = `${LmsApiService.enterpriseAPIBaseUrl}enterprise_customer_catalog/`;
 
   static fetchEnterpriseCatalogQueries = () => LmsApiService.apiClient().get(LmsApiService.enterpriseCatalogQueriesUrl);
+
+  static postEnterpriseCustomerCatalog = (enterpriseCustomerUUID, catalogQueryUUID, title) => LmsApiService
+    .apiClient().post(LmsApiService.enterpriseCustomerCatalogsUrl, {
```
pwnage101: Nit: is it possible to put LmsApiService on the next line?

src/Configuration/Provisioning/ProvisioningForm/PolicyForm/ProvisioningFormCatalog.jsx
```diff
@@ -46,14 +54,14 @@ const ProvisioningFormCatalog = ({ index }) => {
...
-        destination="https://www.google.com"
-      >
-        {formData.policies[index]?.catalogQueryTitle.split(' account')[0]}
-      </Hyperlink>
-      )}
+      {multipleFunds
+        && (
+        <Container>
+          <h4>
+            {formData.policies[index]?.catalogQueryTitle.split(' account')[0]}
```
pwnage101: It feels like this parsing of the catalogQueryTitle is too magical without at least a comment explaining why this is done, or a DRY function since it's repeated in 2 other files.

src/Configuration/Provisioning/ProvisioningForm/ProvisioningForm.jsx
```diff
@@ -12,15 +13,26 @@ import { selectProvisioningContext } from '../data/utils';
 
 const ProvisioningForm = () => {
   const { FORM } = PROVISIONING_PAGE_TEXT;
-  const [multipleFunds, alertMessage] = selectProvisioningContext('multipleFunds', 'alertMessage');
-  // TODO: Extract catalog queries from API to iterate and render policies instead of this for V1
+  const [multipleFunds, alertMessage] = selectProvisioningContext('multipleFunds', 'alertMessage', 'formData');
```
pwnage101: Does this do anything if you don't add a 3rd variable to the left hand side?

src/Configuration/Provisioning/data/tests/utils.test.js
```diff
@@ -36,3 +45,42 @@ describe('lmsCustomerCatalog', () => {
...
+  getAuthenticatedHttpClient: () => ({
+    post: () => Promise.resolve(sampleCreateCatalogResponse),
+  }),
+}));
+describe('createCatalogs', () => {
+  it('returns the correct data', async () => {
+    const data = await createCatalogs([
+      'abf9f43b-1872-4c26-a2e6-1598fc57fbdd',
+      10,
+      'abf9f43b-1872-4c26-a2e6-1598fc57fbdd' - 'title',
```
pwnage101: is this valid syntax?  Are you trying to concatenate two strings?

src/Configuration/Provisioning/ProvisioningForm/PolicyForm/ProvisioningFormCatalog.jsx
```diff
@@ -46,14 +54,15 @@ const ProvisioningFormCatalog = ({ index }) => {
...
-      >
-        {formData.policies[index]?.catalogQueryTitle.split(' account')[0]}
-      </Hyperlink>
-      )}
+      {multipleFunds
+        && (
+        <Container>
+          <h4>
+            {extractDefinedCatalogTitle(formData.policies[index])}
+            {/* {formData.policies[index]?.catalogQueryTitle.split(' account')[0]} */}
```
pwnage101: Leftover comment?

brobro10000: WOOPS! 😅

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormSubmissionButton.jsx
```diff
@@ -1,40 +1,102 @@
...
+  // eslint-disable-next-line consistent-return
+  const handleSubmit = async () => {
+    setSubmitButtonState('pending');
+    // handle subsidy data
+    // handle per policy catalog data
+    if (policies.length === 0 || !canCreatePolicyAndSubsidy) {
+      return setSubmitButtonState('error');
+    }
+    try {
+      /* istanbul ignore next */
```
pwnage101: What's with all these ignores?

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormSubmissionButton.jsx
```diff
@@ -1,40 +1,102 @@
...
+        if (!policy.customerCatalogUUID) {
+          /* istanbul ignore next */
+          const payload = [
+            formData.enterpriseUUID,
+            policy.catalogQueryMetadata.catalogQuery.id,
+            `${formData.enterpriseUUID} - ${policy.catalogQueryMetadata.catalogQuery.title}`,
+          ];
+          /* istanbul ignore next */
+          const catalogCreatedResponse = await createCatalogs(payload);
+          // attach catalogs to policies here
```
pwnage101: Not sure I follow this comment, where exactly is anything getting attached?  Is this just a placeholder?

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormSubmissionButton.jsx
```diff
@@ -1,40 +1,102 @@
...
+      clearFormAndRedirect();
+    }
+  }, [submitButtonState]);
+
+  const buttonLabels = {
+    default: BUTTON.submit,
+    pending: BUTTON.pending,
+    complete: BUTTON.success,
+    error: BUTTON.error,
   };
```
pwnage101: Why make the BUTTON keys different than the buttonLabels keys?

brobro10000: The button label keys are predefined for the component to recognize those value. The `BUTTON` object keys are to reflect the value of the message.

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormSubmissionButton.jsx
brobro10000: Just a placeholder for now, I refined the message as a TODO to be more clear. But when the newly created catalogs are created, they will be associated to the policy api. This would be the potential location of the call.

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormSubmissionButton.jsx
brobro10000: The instanbul next comment is there to temporarily to stifle to codecov warning in Github. We would have to go back and remove these once coverage is increased on the component

pwnage101 (approved):
LGTM! I suggest waiting for another review from somebody more versed in frontend, but I gave this my best shot!

src/Configuration/Provisioning/ProvisioningForm/PolicyForm/ProvisioningFormCatalog.jsx
```diff
@@ -46,14 +54,14 @@ const ProvisioningFormCatalog = ({ index }) => {
...
-      </Hyperlink>
-      )}
+      {multipleFunds
+        && (
+        <Container>
+          <h4>
+            {extractDefinedCatalogTitle(formData.policies[index])}
+          </h4>
+        </Container>
+        )}
```
adamstankiewicz: nit: indentation is a bit off here. also usually prefer to have the `&&` on the same line as the conditional where possible.

src/Configuration/Provisioning/ProvisioningForm/PolicyForm/ProvisioningFormCatalog.jsx
```diff
@@ -46,14 +54,14 @@ const ProvisioningFormCatalog = ({ index }) => {
...
-      <Hyperlink
-        target="_blank"
-        destination="https://www.google.com"
-      >
-        {formData.policies[index]?.catalogQueryTitle.split(' account')[0]}
-      </Hyperlink>
-      )}
+      {multipleFunds
+        && (
+        <Container>
```
adamstankiewicz: [curious/clarification] Is this container necessary? Looking at where another sibling component renders (I think) in its parent ([source](https://github.com/openedx/frontend-app-support-tools/blob/abec0340ac3a781284ba3e500db04b34aaa2b01e/src/Configuration/Provisioning/ProvisioningForm/PolicyForm/ProvisioningFormAccountDetails.jsx#L43)), it appears the sibling component doesn't include a `Container`. Because this `Container` doesn't specifying a `size`, it's fluid and takes up 100% width of its parent.

src/Configuration/Provisioning/ProvisioningForm/ProvisioningForm.jsx
```diff
@@ -13,14 +14,25 @@ import { selectProvisioningContext } from '../data/utils';
 const ProvisioningForm = () => {
   const { FORM } = PROVISIONING_PAGE_TEXT;
   const [multipleFunds, alertMessage] = selectProvisioningContext('multipleFunds', 'alertMessage');
-  // TODO: Extract catalog queries from API to iterate and render policies instead of this for V1
+  const camelCasedQueries = camelCaseObject(getConfig().PREDEFINED_CATALOG_QUERIES);
```
adamstankiewicz: nit: It looks like this logic of camelcasing `getConfig().PREDEFINED_CATALOG_QUERIES` is used in two components. Do you think it's worth abstracting out to a utility function somewhere? Food for thought.

src/Configuration/Provisioning/ProvisioningForm/ProvisioningForm.jsx
```diff
@@ -13,14 +14,25 @@ import { selectProvisioningContext } from '../data/utils';
...
+      },
+    },
+  }));
+
+  const definedCatalogQueries = multipleFunds ? assignedMultipleQueries : defaultQuery;
+
   useEffect(() => {
     resetPolicies();
-    instantiateMultipleFormData(sampleCatalogQuery);
+    instantiateMultipleFormData(definedCatalogQueries);
```
adamstankiewicz: Oof, heads up... this repo has the `react-hooks/exhaustive-deps` ESLint rules turned off ([source](https://github.com/openedx/frontend-app-support-tools/blob/abec0340ac3a781284ba3e500db04b34aaa2b01e/.eslintrc.js#LL13C1-L13C1)). So, you're not getting ESLint warnings around misuse of `useEffect` dependency arrays.

This effect will only run when `multipleFunds` changes given that's what is in its dependency array. However, the effect callback does not rely on `multipleFunds`; it relies on `definedCatalogQueries`. This `useEffect` will not run when `definedCatalogQueries` changes.

src/Configuration/Provisioning/ProvisioningForm/ProvisioningForm.jsx
```diff
@@ -32,14 +44,17 @@ const ProvisioningForm = () => {
       <ProvisioningFormTerm />
       <ProvisioningFormSubsidy />
       <ProvisioningFormAccountType />
-      {!alertMessage && sampleCatalogQuery && sampleCatalogQuery.map(({ uuid, catalogQueryTitle }, index) => (
+      {!alertMessage && definedCatalogQueries && definedCatalogQueries.map(({
```
adamstankiewicz: nit/suggestion: `{!alertMessage && definedCatalogQueries?.map(({`

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormSubmissionButton.jsx
adamstankiewicz: [clarification] @brobro10000 Do you have plans to come back to add coverage for these ignored areas as a fast follow or in an upcoming ticket? This logic is definitely something we'd want to have coverage on so just hoping to clarify what the plan is for that :)

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormSubmissionButton.jsx
```diff
@@ -1,40 +1,101 @@
...
+            formData.enterpriseUUID,
+            policy.catalogQueryMetadata.catalogQuery.id,
+            `${formData.enterpriseUUID} - ${policy.catalogQueryMetadata.catalogQuery.title}`,
+          ];
+          /* istanbul ignore next */
+          const catalogCreatedResponse = await createCatalogs(payload);
+          // TODO: attach newly created catalogs to policies here
+          if (catalogCreatedResponse) {
+            /* istanbul ignore next */
+            return setSubmitButtonState('complete');
```
adamstankiewicz: [clarification] What are the `return` statements being used for here and below? Usually you'd want to be returning during iteration during a `.map`, not a `.forEach`? I'm also not sure a state setter function returns a value?

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormSubmissionButton.jsx
```diff
@@ -1,40 +1,101 @@
...
+        return setSubmitButtonState('error');
+      });
+    } catch (error) {
+      logError(error);
+      setSubmitButtonState('error');
+      /* istanbul ignore next */
+      const { customAttributes } = error;
+      if (customAttributes) {
+        /* istanbul ignore next */
+        return logError(ALERTS.API_ERROR_MESSAGES.ENTERPRISE_CUSTOMER_CATALOG[customAttributes.httpErrorStatus]);
```
adamstankiewicz: Similarly, does `logError` return a value?

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormSubmissionButton.jsx
adamstankiewicz: [clarification] Can you help me understand this logic a bit more? If I'm understanding it correctly, it appears that for each policy, we're calling `createCatalogs` and updating the state of the submit button.

Say we have 2 `policies`. We call `createCatalogs` for the first policy and it does not return a response, the submit button would be up in an `error` state, but then the iteration through the second policy begins, calling `createCatalogs` for it, which does return a response and sets the button state to `complete`.

Does this mean the user would briefly the button say "Failed to create plan. Please try again." before it automatically switches to say `Plan successfully created`, effectively overriding the messaging for the policy that resulted in an error?

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormTerm.jsx
```diff
@@ -36,9 +43,21 @@ const ProvisioningFormTerm = () => {
 
   return (
     <article className="mt-4.5">
-      <div className="mb-1">
+      <Stack direction="horizontal" className="mb-1">
```
adamstankiewicz, line 44: [curious] Why use `Stack` here when it will always only has 1 child? `Stack` is intended to providing equal spacing between elements, but it doesn't really do much with only 1 child.

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormTerm.jsx
```diff
@@ -36,9 +43,21 @@ const ProvisioningFormTerm = () => {
...
+          placement="right"
+          overlay={(
+            <Tooltip
+              id="term-tooltip"
+            >
+              {TERM.TOOLTIP}
+            </Tooltip>
+          )}
+        >
+          <Icon src={InfoOutline} size="sm" className="align-self-baseline ml-1.5" />
```
adamstankiewicz: [clarification/suggestion] Should this be an `IconButton` rather than a standalone `Icon`?

src/Configuration/Provisioning/data/utils.js
```diff
@@ -22,7 +23,11 @@ export const lmsCustomerCatalog = {
   },
 };
 
-// Takes an array of catalog queries and sorted them by last modified date (newest first)
+/**
```
adamstankiewicz, line 27: 🔥 Thanks for adding JSDocs :)

src/Configuration/Provisioning/data/utils.js
```diff
@@ -64,3 +69,59 @@ export function updatePolicies(data, newDataAttribute, index) {
...
   return [...policies];
 }
+
+/**
+ * Checks all form data to ensure that all required fields are filled out,
+ * but not the individual validity of each field.
+ * @param {Object} formData - The form data object.
+ * @returns {Boolean} - Returns true if all form data is valid, false otherwise.
+ */
+export function validFormData(formData) {
```
adamstankiewicz: nit: perhaps `hasValidFormData` or `validateFormData`?

src/Configuration/Provisioning/data/utils.js
```diff
@@ -64,3 +69,59 @@ export function updatePolicies(data, newDataAttribute, index) {
...
+
+    return isAccountNameValid && isAccountValueValid && isCatalogQueryValid && isPerLearnerCapValid;
+  });
+  return isSubsidyValid && arePoliciesValid;
+}
+
+/**
+ *
+ * @param {Array} payload - Array of data to be sent to the LMS, in the form of
+ * [enterpriseCustomerUUID, catalogQueryUUID, title]
```
adamstankiewicz: [suggestion] Having to pass an array of data with the elements in the right order might not be as intuitive as it could be. Any reason why this function couldn't accept the payload as attributes on an object? E.g., `createCatalog({ enterpriseCustomerUUID, catalogQueryUUID, title })`

src/Configuration/Provisioning/data/utils.js
```diff
@@ -64,3 +69,59 @@ export function updatePolicies(data, newDataAttribute, index) {
...
+/**
+ * Extracts the catalog title from the catalogQueryTitle field of a policy.
+ * @param {Object} policy - The policy object.
+ * @returns {String} - The catalog title.
+ */
+export function extractDefinedCatalogTitle(policy) {
+  if (!policy) {
+    return null;
+  }
+  return policy.catalogQueryTitle.split(' account')[0];
```
adamstankiewicz: nit: might be helpful to add a comment on why we're splitting on " account" specifically to add some context for future us.

src/Configuration/Provisioning/ProvisioningForm/PolicyForm/ProvisioningFormCatalog.jsx
brobro10000: Looking at the figma mocks this particular component is not indented. Removed

src/Configuration/Provisioning/ProvisioningForm/ProvisioningForm.jsx
brobro10000: Adding the `definedCatalogQueries` as a dependency results in infinite rendering on the 2 policy state, so I purposely left it off. Since `multipleFunds` value changes when the user selects single vs multiple policy and definedCatalogQueries values are determined by the state of multiple funds, I think its appropriate to leave a single dependency of multiple funds.

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormTerm.jsx
brobro10000: Using a `Stack` component have always been a quick way to get a pure horizontal row effect using a Paragon component.
What I mean by 'pure' is a component that adds a horizontal effect without any additional padding/margin.
The `.row` utility class has a built in -16 margin on either side that offsets the value which requires the additional step of overriding the value with a `ml-0` in our use case.
I was attempting to use `Stack` as a way to serve our use case with built in Paragon components :)

But since has been refactored to the following 
Screenshot 2023-05-02 at 10 42 13 AM

src/supportHeader/DesktopHeader.jsx
```diff
@@ -52,7 +53,7 @@ export default class DesktopHeader extends React.Component {
 
     return (
       <Dropdown>
-        <Dropdown.Toggle as={AvatarButton} src={avatar}>
+        <Dropdown.Toggle id={uuidv4()} as={AvatarButton} src={avatar}>
```
brobro10000: Added to remove dropdown requires id error

src/index.jsx
```diff
@@ -50,7 +67,7 @@ subscribe(APP_READY, () => {
         <Header />
         <Switch>
           {/* Start: Configuration Dropdown Routes */}
-          {getConfig().FEATURE_CONFIGURATION_MANAGEMENT && configurationRoutes.map((route) => route)}
+          {getConfig().FEATURE_CONFIGURATION_MANAGEMENT && configurationRoutes}
```
brobro10000, line 70: Refactored to remove Switch error

src/index.jsx
```diff
@@ -38,10 +39,26 @@ subscribe(APP_READY, () => {
     return;
   }
   const configurationRoutes = [
-    <Route path={CONFIGURATION.SUB_DIRECTORY.PROVISIONING.SUB_DIRECTORY.EDIT} component={ProvisioningFormContainer} />,
-    <Route path={CONFIGURATION.SUB_DIRECTORY.PROVISIONING.SUB_DIRECTORY.NEW} component={ProvisioningFormContainer} />,
-    <Route path={CONFIGURATION.SUB_DIRECTORY.PROVISIONING.HOME} component={ProvisioningPage} />,
-    <Route path={CONFIGURATION.HOME} component={ConfigurationPage} />,
+    <Route
+      key={uuidv4()}
```
brobro10000, line 43: Added keys to remove missing key error

src/Configuration/Provisioning/ProvisioningForm/ProvisioningForm.jsx
adamstankiewicz: If the disabled ESLint is turned on again (which it really should be), ESLint will start complaining about this `useEffect` with `multipleFunds` as the dependency array. The infinite rendering is occurring because `definedCatalogQueries` is a new array/object reference on each re-render. You could likely memoize `definedCatalogQueries` via a `useMemo` to get around that.

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormTerm.jsx
adamstankiewicz: Gah, I'm sorry... I totally missed that there was an `h3` in here, too... I had thought it was just the `OverlayTrigger` as the children for `Stack`, which led to the original comment.

I think your prior use of `Stack` was correct, given you're trying to provide a gap between the `h3` and the `OverlayTrigger` (now `IconButtonWithTooltip`).

Apologies for the confusion, Hamzah!

src/supportHeader/DesktopHeader.jsx
adamstankiewicz: nit: might recommend just using a string similar to "desktop-header-dropdown-toggle" rather than needing to generate a UUID on each re-render.

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormSubmissionButton.jsx
brobro10000: Refactored to check if a `uuid` is returned from catalog creation, but this may not even be necessary. The `catch` in the try catch kicks in when anything other then non-successful attempt to create a catalog is done. Furthermore, the `hasValidData` function checks each policy to make sure that each field is populated.

Subsequent PRs will check against existing customers to see if the `enterpriseUUID` passed is a valid uuid for an existing customer along with the additional validations previously implemented for data format. 

I was able to catch a use case and better architect the API call for predefined catalogs in the refactor.

brobro10000: I software engineered better in v2 😅

src/Configuration/Provisioning/ProvisioningForm/ProvisioningForm.jsx
```diff
@@ -8,38 +8,58 @@ import ProvisioningFormPolicyContainer from './PolicyForm';
...
+      },
+    },
+  }));
+
+  const definedCatalogQueries = useMemo(() => {
+    if (multipleFunds) {
+      return assignedMultipleQueries;
+    }
+    return defaultQuery;
+  }, [multipleFunds]);
```
adamstankiewicz: All values used within the `useMemo` callback function should be part of the dependency array; otherwise, `definedCatalogQueries` won't be updated if/when `assignedMultipleQueries` or `defaultQuery` changes.

Note that by adding `assignedMultipleQueries` to the dependency array as is would result in the `useMemo` always triggering due to the new array reference for `assignedMultipleQueries` on each re-render. This could also be memoized to avoid the dependency array issues.

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormSubmissionButton.jsx
```diff
@@ -1,40 +1,87 @@
...
+  const { policies } = formData;
+  const canCreatePolicyAndSubsidy = hasValidData(formData);
+
+  const [submitButtonState, setSubmitButtonState] = useState('default');
+
+  const clearFormAndRedirect = () => {
     resetFormData();
     history.push(HOME);
   };
+  // eslint-disable-next-line consistent-return
```
adamstankiewicz: nit: avoid disabling this rule

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormSubmissionButton.jsx
```diff
@@ -1,40 +1,87 @@
+// TODO: Remove instanbul next and test file when submission button is complete
```
adamstankiewicz: nit: is this comment still relevant?

src/Configuration/Provisioning/ProvisioningForm/ProvisioningForm.jsx
```diff
@@ -8,38 +8,58 @@ import ProvisioningFormPolicyContainer from './PolicyForm';
...
           key={uuid}
           title={catalogQueryTitle}
           index={index}
         />
       ))}
-      {(multipleFunds === undefined || alertMessage) && (
+      {alertMessage && (
       <Alert variant="warning" className="mt-5">
         {alertMessage}
       </Alert>
```
adamstankiewicz: nit: indent

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormSubmissionButton.jsx
```diff
@@ -1,40 +1,87 @@
...
     resetFormData();
     history.push(HOME);
   };
+  // eslint-disable-next-line consistent-return
+  const handleSubmit = async () => {
+    setSubmitButtonState('pending');
+    // handle subsidy data
+    // handle per policy catalog data
+    if (policies.length === 0 || !canCreatePolicyAndSubsidy) {
+      return setSubmitButtonState('error');
```
adamstankiewicz: Similar feedback as https://github.com/openedx/frontend-app-support-tools/pull/328#discussion_r1182480066 from last week about returning state setter functions.

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormTerm.jsx
```diff
@@ -36,9 +41,19 @@ const ProvisioningFormTerm = () => {
 
   return (
     <article className="mt-4.5">
-      <div className="mb-1">
-        <h3>{TERM.TITLE}</h3>
-      </div>
+      <Stack direction="horizontal" className=" mb-1">
```
adamstankiewicz: nit: unnecessary space before "mb-1"

src/Configuration/Provisioning/data/constants.js
```diff
@@ -12,7 +12,10 @@ const PROVISIONING_PAGE_TEXT = {
     TITLE: (pathName) => `${titleCase(pathName.split('/').reverse()[0])} Learner Credit Plan`,
     SUB_TITLE: 'Plan Details',
     BUTTON: {
-      submit: 'Create Learner Credit Plan',
+      submit: 'Create learner credit plan',
+      pending: 'Creating...',
+      success: 'Plan successfully created',
+      error: 'Failed to create plan. Please try again.',
```
adamstankiewicz, line 18: [curious] How well does this longer text in the button (e.g., does the button become awkwardly long and/or its text wrap?)

src/Configuration/Provisioning/data/utils.js
```diff
@@ -64,3 +70,89 @@ export function updatePolicies(data, newDataAttribute, index) {
...
   return [...policies];
 }
+
+/**
+ * Checks all form data to ensure that all required fields are filled out,
+ * but not the individual validity of each field.
+ * @param {Object} formData - The form data object.
+ * @returns {Boolean} - Returns true if all form data is valid, false otherwise.
+ */
+export function hasValidData(formData) {
```
adamstankiewicz: nit: perhaps name this function a bit more explicitly, e.g. `hasValidPolicyAndSubidy`?

src/Configuration/Provisioning/data/constants.js
brobro10000: Appearance of button during failed state. The button is rendering normally, and appears normal in size relative to the form despite the longer text 👍🏽 

Screenshot 2023-05-08 at 12 47 32 PM

src/Configuration/Provisioning/ProvisioningForm/ProvisioningForm.jsx
brobro10000: This component ended up getting removed due to guidance from Jenn 👍🏽

src/Configuration/Provisioning/ProvisioningForm/ProvisioningForm.jsx
brobro10000: Ended up moving this logic over to the hooks file. Updating the dependency array in the main `ProvisioningForm` file ended up causing `hasValidPolicyAndSubidy` to fail because of all the re-rendering of the values resulting in the policies array to clear all existing data.

PR #329, opened 2023-04-18
feat: allow only whole dollar amounts in inputs for money

ghost:
```
commit d34fffa4f9a5de6a2c52a73aa639bfdec48b9559
Author: Emily Aquin <eaquin@2u.com>
Date:   Tue Apr 18 10:28:28 2023 -0500


In the Support Tools: New Learner Credit Plan, a validity check was added that the 
"Total account value" and the "Per learner spend cap" can only be numbers, and 
no decimal point. If the user types any character other than a number, 
the account value or the spend cap will not update on the screen.

ENT-6940
```
A screenshot of the error if the user has just typed a letter or `.`(note that the incorrect character does not show in the input)
Screenshot 2023-04-18 at 10 56 40 AM

### DIFF
--- a/src/Configuration/Provisioning/ProvisioningForm/PolicyForm/ProvisioningFormAccountDetails.jsx
+++ b/src/Configuration/Provisioning/ProvisioningForm/PolicyForm/ProvisioningFormAccountDetails.jsx
@@ -3,16 +3,18 @@ import { Form } from '@edx/paragon';
 import PROVISIONING_PAGE_TEXT from '../../data/constants';
 import useProvisioningContext from '../../data/hooks';
 import { indexOnlyPropType, selectProvisioningContext } from '../../data/utils';
+import { isWholeDollarAmount } from '../../../../utils';
 
 const ProvisioningFormAccountDetails = ({ index }) => {
-  const { ACCOUNT_DETAIL } = PROVISIONING_PAGE_TEXT.FORM;
+  const { ACCOUNT_DETAIL, ALERTS } = PROVISIONING_PAGE_TEXT.FORM;
   const { setAccountName, setAccountValue } = useProvisioningContext();
   const [multipleFunds, formData] = selectProvisioningContext('multipleFunds', 'formData');
   const formFeedbackText = multipleFunds
     ? ACCOUNT_DETAIL.OPTIONS.totalAccountValue.dynamicSubtitle(formData.policies[index]?.catalogQueryTitle.split(' account')[0])
     : ACCOUNT_DETAIL.OPTIONS.totalAccountValue.subtitle;
-  const [accountValueState, setAccountValueState] = useState(null);
-  const [accountNameState, setAccountNameState] = useState(null);
+  const [accountValueState, setAccountValueState] = useState('');
+  const [accountNameState, setAccountNameState] = useState('');
+  const [isWholeDollar, setIsWholeDollar] = useState(true);
 
   const handleChange = useCallback((e) => {
     const newEvent = e.target;
@@ -20,7 +22,14 @@ const ProvisioningFormAccountDetails = ({ index }) => {
     if (dataset.testid === 'account-name') {
       setAccountName({ accountName: value }, index);
       setAccountNameState(value);
-    } else if (dataset.testid === 'account-value') {
+      return;
+    }
+    if (dataset.testid === 'account-value') {
+      if (value !== '' && !isWholeDollarAmount(value)) {
+        setIsWholeDollar(false);
+        return;
+      }
+      setIsWholeDollar(true);
       setAccountValue({ accountValue: value }, index);
       setAccountValueState(value);
     }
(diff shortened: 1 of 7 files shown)


src/Configuration/Provisioning/ProvisioningForm/PolicyForm/ProvisioningFormAccountDetails.jsx
```diff
@@ -49,6 +56,13 @@ const ProvisioningFormAccountDetails = ({ index }) => {
         <Form.Control.Feedback>
           {formFeedbackText}
         </Form.Control.Feedback>
+        {!isWholeDollar && (
+          <Form.Control.Feedback
+            type="invalid"
+          >
+            { ALERTS.incorrectDollarAmount }
```
adamstankiewicz: nit: spaces before/after the `{` and `}` unnecessary.

src/utils/index.js
```diff
@@ -4,6 +4,7 @@ const COURSE_ID_REGEX = '[^/+]+(/|\\+)[^/+]+(/|\\+)[^/?]+';
 const EMAIL_REGEX = '^[a-zA-Z0-9\'!#$&*._%+-]+@[a-zA-Z0-9.-]+\\.[a-zA-Z]{2,}$';
 const USERNAME_REGEX = '^[\\w.@_+-]+$';
 const LMS_USER_ID_REGEX = '^(?!0)[0-9]+$';
+const DIGITS_ONLY_REGEX = /^\d+$/;
```
brobro10000: I would update this regex to take into account 0 for leading values 
`/^[1-9](?:\d+)?&/`. We can verify with Joe 

The ?: before the \d+ is basically a negation flag to tell the regex not to remember the values in the parenthesis. Added to improve performance of the regex.  The  ? after the (?:\d+) is an optional flag to ignore these values if they don't exist allowing the regex to complete.
Screenshot 2023-04-19 at 11 11 37 AM

brobro10000 (approved):
Contingent approval till we get clarification on the 0 value for learner spend cap value and total account value.  

Approving in the mean time.  Great job! 👍🏽

src/utils/index.js
ghost: Thanks for checking! I'll make the change

brobro10000 (approved):
LGTM, Great job.

src/Configuration/Provisioning/ProvisioningForm/PolicyForm/ProvisioningFormAccountDetails.jsx
```diff
@@ -3,24 +3,33 @@ import { Form } from '@edx/paragon';
...
     const newEvent = e.target;
     const { value, dataset } = newEvent;
     if (dataset.testid === 'account-name') {
       setAccountName({ accountName: value }, index);
       setAccountNameState(value);
-    } else if (dataset.testid === 'account-value') {
+      return;
+    }
+    if (dataset.testid === 'account-value') {
+      if (value && !isWholeDollarAmount(value)) {
```
adamstankiewicz: [nit/clarification] Is the `value &&` check here (and elsewhere) necessary, or could we rely on just the logic in `isWholeDollarAmount` which seems to include a check for `value &&` already?

src/Configuration/Provisioning/ProvisioningForm/PolicyForm/ProvisioningFormPerLearnerCapAmount.jsx
```diff
@@ -1,19 +1,29 @@
...
 const ProvisioningFormPerLearnerCapAmount = ({ index }) => {
-  const { LEARNER_CAP_DETAIL } = PROVISIONING_PAGE_TEXT.FORM;
+  const { LEARNER_CAP_DETAIL, ALERTS } = PROVISIONING_PAGE_TEXT.FORM;
   const { setPerLearnerCap } = useProvisioningContext();
+  const [isWholeDollar, setIsWholeDollar] = useState(true);
+  const [perLearnerCapValue, setPerLearnerCapValue] = useState('');
 
   const handleChange = (e) => {
     const newEventValue = e.target.value;
+    if (newEventValue && !isWholeDollarAmount(newEventValue)) {
```
adamstankiewicz: [nit/clarification] Is the `newEventValue &&` check here (and elsewhere) necessary, or could we rely on just the logic in `isWholeDollarAmount` which seems to include a check for `value &&` already?

adamstankiewicz (approved):
LGTM with some nits.

src/Configuration/Provisioning/ProvisioningForm/PolicyForm/ProvisioningFormAccountDetails.jsx
ghost: I tried it out without the `value &&` check, and it seems that it won't allow the user to completely backspace out their input if I do not have the `value &&`. I think it's because value is conditionally rendering text.
There's a few different ways I could solve this, but I decided on being explicit and checking that it isn't in the original state (an empty string).

src/Configuration/Provisioning/ProvisioningForm/PolicyForm/ProvisioningFormPerLearnerCapAmount.jsx
ghost: Same comment as above:
I tried it out without the value && check, and it seems that it won't allow the user to completely backspace out their input if I do not have the value &&. I think it's because value is conditionally rendering text.
There's a few different ways I could solve this, but I decided on being explicit and checking that it isn't in the original state (an empty string).

PR #331, opened 2023-04-26
feat: validates UUID to an existing enterprise customer

brobro10000:
Implements the 'basic_list' enterprise customer API endpoint to dynamically create a dropdown based on the user's inputted value. Also makes use of `debounce` to prevent query on key down, and delays api call to 1 second after the final keystroke is complete.


src/data/services/EnterpriseApiService.js
```diff
@@ -8,7 +8,11 @@ class LmsApiService {
 
   static enterpriseCatalogQueriesUrl = `${LmsApiService.baseUrl}/enterprise/api/v1/enterprise_catalog_query/`;
 
+  static enterpriseCustomersBasicListUrl = `${LmsApiService.baseUrl}/enterprise/api/v1/enterprise-customer/basic_list/`;
+
   static fetchEnterpriseCatalogQueries = () => LmsApiService.apiClient().get(LmsApiService.enterpriseCatalogQueriesUrl);
+
+  static fetchEnterpriseCustomersBasicList = (query) => LmsApiService.apiClient().get(`${LmsApiService.enterpriseCustomersBasicListUrl}${query !== undefined ? `?q=${query}` : ''}`);
```
pwnage101: ditto https://github.com/openedx/edx-enterprise/pull/1742#pullrequestreview-1404735552

src/data/services/EnterpriseApiService.js
```diff
@@ -8,7 +8,11 @@ class LmsApiService {
 
   static enterpriseCatalogQueriesUrl = `${LmsApiService.baseUrl}/enterprise/api/v1/enterprise_catalog_query/`;
 
+  static enterpriseCustomersBasicListUrl = `${LmsApiService.baseUrl}/enterprise/api/v1/enterprise-customer/basic_list/`;
+
   static fetchEnterpriseCatalogQueries = () => LmsApiService.apiClient().get(LmsApiService.enterpriseCatalogQueriesUrl);
+
+  static fetchEnterpriseCustomersBasicList = (query) => LmsApiService.apiClient().get(`${LmsApiService.enterpriseCustomersBasicListUrl}${query !== undefined ? `?name_or_uuid=${query}` : ''}`);
```
pwnage101: Nit: also rename the function argument "query" -> "enterprise_name_or_uuid" to avoid all confusion

pwnage101 (approved):

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormCustomerDropdown.jsx
```diff
@@ -0,0 +1,68 @@
...
+  const { setCustomerUUID, getCustomers } = useProvisioningContext();
+  const [selected, setSelected] = useState({ title: '' });
+  const [dropdownValues, setDropdownValues] = useState(['No matching enterprise']);
+
+  const debouncedSearch = useMemo(() => debounce(getCustomers, 500, {
+    leading: false,
+  }), [formData.enterpriseUUID]);
+  const handleOnSelected = (value) => {
+    if (value && value.includes('---')) {
+      const valueUuid = value.split(' --- ')[1].trim();
```
adamstankiewicz, line 26: nit: Is it worth having a comment with a brief explanation for why this `.includes('---')` and `.split(' --- ')` is necessary?

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormCustomerDropdown.jsx
```diff
@@ -0,0 +1,68 @@
...
+  };
+  const updateDropdown = (value) => {
+    setCustomerUUID(value);
+    if (value?.length === 0 || customers.length === 0) {
+      return setDropdownValues(['No matching enterprise']);
+    }
+    return updateDropdownList();
+  };
+  useEffect(() => {
+    const delayDebounceFn = setTimeout(() => {
```
adamstankiewicz, line 45: [clarification] Can you explain why this `setTimeout` is necessary?

src/data/services/EnterpriseApiService.js
adamstankiewicz: Bump.

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormCustomerDropdown.jsx
```diff
@@ -0,0 +1,68 @@
...
+      debouncedSearch(formData.enterpriseUUID);
+    }, 500);
+    return () => clearTimeout(delayDebounceFn);
+  }, [formData.enterpriseUUID]);
```
adamstankiewicz: `debouncedSearch` is not in `useEffect` dependency array. With it added, it _may_ resolve the need for the `setTimeout` above.

brobro10000: Did not seem to fix the debounce issue I was facing :(

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormCustomerDropdown.jsx
brobro10000: The setTimeout is the time between the onChange effect from the users interaction (typing a uuid/customer name).
Not adding that in (even with the updated dependency array results in a call of debouncedSearch every key stroke after the delay that was set within the debounce function. The timeout is acting as a buffer before the the function is called.

PR #333, opened 2023-05-08
feat: Add datatable skeleton for enterprise subsidy list

brobro10000:
https://2u-internal.atlassian.net/browse/ENT-6941
Adds a datatable skeleton to the main dashboard of the provisioning page.

To be populated with existing subsidies once API implemented.
Handles partial routing with the exception of populating edit page with appropriate values.
Uses DataTable built in search bar functionality.

Given the current spec of subsidy, the accessors of the table have been updated to match the endpoint fields, along with data transformation on the UUID mapping, and the date time from Zulu UTC to local dashified date string
These changes handle this tickets AC: https://2u-internal.atlassian.net/browse/ENT-6943

Tests will be deferred until API integration is complete. Its its current state, a majority of the page is currently tested since it just renders a data table with test data. 

Screenshot 2023-05-16 at 10 27 58 AM


pwnage101:
Should we design for the possibility that the customer has two learner credit plans (i.e. two subsidies)?

src/Configuration/Provisioning/data/constants.js
```diff
@@ -1,6 +1,9 @@
 import { v4 as uuidv4 } from 'uuid';
 import { titleCase } from '../../../utils';
 
+// Set to true or false to enable local testing, populates DataTable with sample data
+export const TESTING = false;
```
adamstankiewicz: nit: perhaps use a more explicit name than just `TESTING`? E.g., maybe something like `SHOULD_USE_LOCAL_DATA` or `USES_LOCAL_TEST_DATA`

src/Configuration/Provisioning/data/utils.js
```diff
@@ -187,6 +187,32 @@ export function getCamelCasedConfigAttribute(attribute) {
...
 }
 
+/**
+ *
+ * @param {Array} fetchedData - The data fetched from the API
+ * @param {Function} actionIcon - The icon to be displayed in the action column, passes the redirectURL function
+ * @param {Function} redirectURL - The function to be called when the icon is clicked, redirects to passed UUID
+ * @returns - The normalized data to be displayed in the table
+ */
+export function normalizeSubsidyDataTableData(fetchedData, actionIcon, redirectURL) {
```
adamstankiewicz: nit: may be helpful to use an object for the function signature, e.g. `normalizeSubsidyDataTableData({ fetchedData, actionIcon, redirectUrl })`. I generally use an object for the args if the function has (or could have) more than 1 argument, so it acts more like keyword arguments (kwargs) without having to rely on passing the function args in a specific order each time you use it.

adamstankiewicz (approved):

PR #336, opened 2023-05-17
feat: Create error page for API errors

brobro10000:
Adds Error page routing and UI to handle provisioning form errors relating to API calls. 
Adds logic in the provisioning form button component to handle and redirect errors to error page, passing the error message and status code to the error page for attempts at creating a customer catalog, creating a subsidy or creating a policy

Added relevant testing. 

Error Page.
Screenshot 2023-05-18 at 9 41 36 AM


src/Configuration/Provisioning/ErrorPage/ErrorPageButton.jsx
```diff
@@ -0,0 +1,21 @@
+import { Button } from '@edx/paragon';
+import PropTypes from 'prop-types';
+
+const ErrorPageButton = ({ buttonInteraction, buttonText }) => (
+  <div className="mt-4">
```
adamstankiewicz: nit: it might be better to leave the spacing concerns up to the caller of this component. For example, will all usages of `ErrorPageButton` always want `.mt-4` for all use cases? In some cases, you may have different spacing needs.

Might recommend relying on a `className` passed by the caller of the component., e.g.:

```jsx
const ErrorPageButton = ({ className, children }) => (
  <Button className={className}>
    {children}
  </Button>
);
```

(note, my removal of the wrapping `div` was intentional, unless it's needed for the layout?)

src/Configuration/Provisioning/ErrorPage/ErrorPageButton.jsx
```diff
@@ -0,0 +1,21 @@
...
+    <Button
+      variant="primary"
+      as="a"
```
adamstankiewicz: From the looks of it, you're trying to treat `Button` as a hyperlink even though you're relying on `history.push` to do the navigation. If a button is a hyperlink, it indeed should get rendered as an `a`, but in a way that takes advantage of React routing, e.g. the `Link` component in `react-router-dom` but it should also have an `href` attribute on the underlying DOM element.

Might recommend instead of passing `buttonInteraction` and `buttonText`, to make this component a bit more flexible and semantic, e.g.:

```jsx
const ErrorPageButton = ({ children, className, as }) => (
  <Button as={as} variant="primary" className={className}>
    {children}
  </Button>
);
```

Then, I think you could do something like:

```jsx
<ErrorPageButton
  as={Link}
  to={ROUTES.CONFIGURATION.SUB_DIRECTORY.PROVISIONING}
  className="mt-4"
>
  {ERROR_PAGE_TEXT.BUTTON}
</ErrorPageButton>
```

Where the button is a true hyperlink (via `Link`). For example, users could `cmd+click` the button to open its destination in a new tab.

src/Configuration/Provisioning/ErrorPage/ErrorPageContainer.jsx
```diff
@@ -0,0 +1,41 @@
...
+  }, [locationState.errorMessage]);
+
+  // Went with this implementation to allow for future expansion of error page
+  const handleOnClick = () => {
+    history.push(HOME);
+  };
+
+  return (
+    <Container size="md" className="mt-5 text-center">
+      <ErrorPageImage image={ErrorPage} imageAltText="Portable computer in need of a repair shop" />
```
adamstankiewicz: nit: Haha, I might argue that the image here is purely decorative, and an `alt` description isn't really necessary. Does this value need to get read to the user for the page to make sense, or would it actually cause more confusion to hear the words "Image, portable computer in need of a repair shop" spoken to you while parsing the page with a screen reader like VoiceOver.

src/Configuration/Provisioning/ErrorPage/ErrorPageImage.jsx
```diff
@@ -0,0 +1,17 @@
...
+  <Image
+    src={image}
+    alt={imageAltText}
+    fluid
+  />
+);
+
+ErrorPageImage.propTypes = {
+  image: PropTypes.string.isRequired,
+  imageAltText: PropTypes.string.isRequired,
```
adamstankiewicz: Related to other comment, `alt` tags may not _always_ be required. Would consider whether these images are purely decorate or provide semantic meaning to the user.

src/Configuration/Provisioning/ErrorPage/ErrorPageMessage.jsx
```diff
@@ -0,0 +1,34 @@
...
+          {ERROR_PAGE_TEXT.SPANNED_TITLE}
+        </span>
+      </h1>
+    </div>
+    <div className="mt-4">
+      <p>
+        {message}
+        <br />
+        {ERROR_PAGE_TEXT.SUB_TITLE}
+      </p>
```
adamstankiewicz: nit: maybe use two distinct `<p>` tags, one for `{message}` and one for `{ERROR_PAGE_TEXT.SUBTITLE}` to remove the need for a `<br />` to separate them. `p` tags have bottom margin baked in.

src/Configuration/Provisioning/ErrorPage/tests/ErrorPageImage.test.jsx
```diff
@@ -0,0 +1,17 @@
...
+describe('<ErrorPageImage />', () => {
+  it('renders the image', () => {
+    const imageAltText = 'Test Image Alt Text';
+    renderWithRouter(
+      <ErrorPageImage
+        image={ErrorPage}
+        imageAltText={imageAltText}
+      />,
+    );
+    expect(screen.getByAltText(imageAltText)).toBeTruthy();
```
adamstankiewicz, line 17: While you can use the alt text to select the image element, you'll also likely want to assert that is has the correct `src`, given the `image` passed in.

src/Configuration/Provisioning/data/constants.js
```diff
@@ -145,34 +145,46 @@ const PROVISIONING_PAGE_TEXT = {
...
           400: 'The subsidy could not be created.',
           401: 'Authentication failed.',
           403: 'Authentication recognized but incorrect credentials.',
           404: 'Subsidy creation endpoint failed to respond',
           405: 'The subsidy could not be created.',
+          500: 'System failure',
         },
         POLICY_CREATION: {
-          400: 'The subsidy could not be created.',
+          400: 'The policy could not be created.',
```
adamstankiewicz, line 171: Nice catches

src/Configuration/Provisioning/ErrorPage/ErrorPageContainer.jsx
adamstankiewicz: Also, did you use AI to generate the image alt text?! 😉

brobro10000: lol nope, completely form the noggin :D

src/Configuration/Provisioning/ErrorPage/data/constants.js
```diff
@@ -0,0 +1,7 @@
+// eslint-disable-next-line import/prefer-default-export
+export const ERROR_PAGE_TEXT = {
+  TITLE: 'We\'re sorry.',
+  SPANNED_TITLE: 'Something went wrong.',
+  SUB_TITLE: 'Please try again later. If the problem persists, please contact your system administrator.',
+  BUTTON: 'Return to Learner Credit Plans',
+};
```
brobro10000, line 7: Ended up moving the constants into the same directory for 'true' extensibility in the future. So the component folder can be dragged and drop to other locations without static dependencies outside the folder (beyond packages)

src/Configuration/Provisioning/ErrorPage/ErrorPageButton.jsx
```diff
@@ -0,0 +1,31 @@
...
+    className={className}
+    onClick={onClick}
+    variant="primary"
+  >
+    {children}
+  </Button>
+);
+
+ErrorPageButton.propTypes = {
+  as: PropTypes.shape({}).isRequired,
```
adamstankiewicz: nit: could use `PropTypes.node` here.

src/Configuration/Provisioning/ErrorPage/ErrorPageButton.jsx
```diff
@@ -0,0 +1,31 @@
...
+  to: PropTypes.string.isRequired,
+  className: PropTypes.string,
+  children: PropTypes.node.isRequired,
+  onClick: PropTypes.func,
+};
+
+ErrorPageButton.defaultProps = {
+  className: '',
```
adamstankiewicz: nit: typically, the default prop for a `className` would be undefined.

adamstankiewicz (approved):
LGTM! just 2 small nits

adamstankiewicz (approved):
LGTM! just 2 small nits

PR #343, opened 2023-06-13
feat: populates datatable with subsidies

brobro10000:
Populates datatable with data from the subsidy endpoint
Also sorts and filters leveraging manualSort/manualFilter

Revamps validation for message and regex logic to handle '00k' prefix for the opportunity product. 
Screenshot 2023-06-27 at 4 10 24 PM

Screenshot 2023-06-27 at 1 58 28 PM

Pending updated tests


src/Configuration/Provisioning/DashboardDatatable.jsx
```diff
@@ -1,34 +1,80 @@
...
+          src={SvgDjango}
+          iconAs={Icon}
+        />
+      </Hyperlink>,
+    ]
+  );
+
+  const [learnerCreditCustomers, setLearnerCreditCustomers] = useState(data?.enterpriseSubsidies || []);
+  const [pageIndex, setPageIndex] = useState(0);
+  const [stateChange, setStateChange] = useState(true);
```
adamstankiewicz: [clarification] What is the intent behind `stateChange`?

src/Configuration/Provisioning/DashboardDatatable.jsx
```diff
@@ -1,34 +1,80 @@
...
-
-  // Implementation due to filterText value displaying accessor value customerName as opposed to Customer Name
-  const filterStatus = (rest) => <DataTable.FilterStatus showFilteredFields={false} {...rest} />;
+  console.log(data.enterpriseSubsidies);
   return (
     <section className="mt-5">
       <DataTable
         isPaginated
         isSortable
         isFilterable
```
adamstankiewicz: I believe you may also want to include `manualSortBy` and `manualFilters`, similar to `manualPagination`?

src/Configuration/Provisioning/DashboardDatatable.jsx
```diff
@@ -1,34 +1,80 @@
...
         defaultColumnValues={{ Filter: TextFilter }}
+        pageCount={data.enterpriseSubsidies.pageCount}
         initialState={{
           pageSize: MAX_PAGE_SIZE,
-          pageIndex: 0,
+          pageIndex,
```
adamstankiewicz: Given this is the _initial table state_, should we be passing a dynamic `pageIndex` that changes as the user paginates the table? Or should this be set as a constant `0` given its pertaining to initial state?

package.json
```diff
@@ -29,10 +29,10 @@
     "url": "https://github.com/openedx/frontend-app-support-tools/issues"
   },
   "dependencies": {
-    "@edx/brand": "npm:@edx/brand-edx.org@^1.4.2",
+    "@edx/brand": "npm:@edx/brand-openedx@1.2.0",
```
brobro10000, line 32: Updated to handle `"~@edx/brand/paragon/overrides"` import for datatable styles

package.json
```diff
@@ -29,10 +29,10 @@
...
     "@edx/frontend-enterprise-utils": "^3.0.0",
     "@edx/frontend-platform": "^4.2.0",
-    "@edx/paragon": "^20.26.0",
+    "@edx/paragon": "20.45.0",
```
brobro10000, line 35: Upgraded to handle new `DjangoShort` icon
Screenshot 2023-06-26 at 11 38 34 AM

src/Configuration/tests/ConfigurationPage.test.jsx
```diff
@@ -1,24 +0,0 @@
-import { screen } from '@testing-library/react';
```
brobro10000, line 1: Removed because of duplicate file

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormCustomer.jsx
```diff
@@ -53,6 +53,11 @@ const ProvisioningFormCustomer = () => {
           onChange={handleChange}
           data-testid="customer-financial-identifier"
         />
+        {isOpportunityProduct && (
+        <Form.Control.Feedback>
+          {financialIdentifier.length}/{CUSTOMER.FINANCIAL_IDENTIFIER.MAX_LENGTH}
```
brobro10000: Added a character counter validation feedback on opportunity product field

package.json
adamstankiewicz: [inform] Additional context: because this is an open-source repo in the `openedx` Github organization, the `brand-openedx` theme should be the default theme installed. Should a consumer want to use the `brand-edx.org` theme for an MFE, that should be configured outside of committed code. For 2U/edX, the `@edx/brand` package may be overridden at build+deploy time via configuration ([example](https://github.com/edx/edx-internal/blob/62c1f4412692ef9da4e50efac2b33a877c83f02b/frontends/frontend-app-profile/prod_config.yml#L13-L14)).

package.json
```diff
@@ -42,7 +42,6 @@
     "classnames": "2.2.6",
     "lodash.debounce": "4.0.8",
     "moment": "2.29.1",
-    "newrelic": "5.13.1",
```
adamstankiewicz, line 45: [inform] Context: `newrelic` is installed, but seemingly unused by this repo. All New Relic related things for MFEs currently comes from `@edx/frontend-build` and `@edx/frontend-platform`, without consumers needing to install `newrelic` themselves.

Having `newrelic` in package.json causes issues such as:

```shell
❯ npm i
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'newrelic@5.13.1',
npm WARN EBADENGINE   required: { node: '>=6.0.0 <13.0.0', npm: '>=3.0.0' },
npm WARN EBADENGINE   current: { node: 'v18.16.0', npm: '9.5.1' }
npm WARN EBADENGINE }
npm WARN deprecated newrelic@5.13.1: This version of the New Relic Node Agent has reached the end of life.
```

src/Configuration/Provisioning/DashboardDatatable/DashboardDatatable.jsx
```diff
@@ -0,0 +1,114 @@
+import {
```
adamstankiewicz: tiny nit: might recommend renaming component name's casing to `DashboardDataTable` vs. `DashboardDatatable` to more closely match the `DataTable` component casing.

src/Configuration/Provisioning/DashboardDatatable/DashboardTableActions.jsx
```diff
@@ -0,0 +1,58 @@
...
+import PropTypes from 'prop-types';
+import ROUTES from '../../../data/constants/routes';
+
+const DashboardTableActions = ({ row }) => {
+  const rowUuid = row.values.uuid;
+  const { DJANGO_ADMIN_SUBSIDY_BASE_URL } = getConfig();
+  const history = useHistory();
+  const { HOME } = ROUTES.CONFIGURATION.SUB_DIRECTORY.PROVISIONING;
+  return [
+    getConfig().FEATURE_CONFIGURATION_EDIT_ENTERPRISE_PROVISION && (
```
adamstankiewicz: I might recommend conditionally pushing to the array rather than using an inline conditional within the array. I believe if `getConfig().FEATURE_CONFIGURATION_EDIT_ENTERPRISE_PROVISION` were to evaluate as a falsey value, `DashboardTableActions` would return an array with that falsey value as its first element.

Example:

```
[false && 'hi']
```

evaluates to `[false]` which is likely not expected here?

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormCustomer.jsx
```diff
@@ -53,6 +53,11 @@ const ProvisioningFormCustomer = () => {
...
+        </Form.Control.Feedback>
```
adamstankiewicz: tiny nit: could use extra indent to match other `<Form.Control.Feedback>` components (e.g., on line 62 below).

src/data/services/SubsidyApiService.js
```diff
@@ -1,9 +1,27 @@
...
   static apiClient = getAuthenticatedHttpClient;
 
+  static getAllSubsidies = ({
+    paginatedURL,
+    pageSize,
+    sortBy,
+    filteredData,
+  }) => {
+    const subsidiesURL = `${getConfig().SUBSIDY_BASE_URL}/api/v1/subsidies/`;
+    let optionalUrlParams = '';
```
adamstankiewicz: Might recommend using `URLSearchParams` here to construct the query params string, rather than using string concatenation.

adamstankiewicz: Would recommend using `URLSearchParams` to create the query params string.

src/data/services/SubsidyApiService.js
```diff
@@ -1,9 +1,27 @@
...
+
+    optionalUrlParams += pageSize ? `&page_size=${pageSize}` : '';
+    optionalUrlParams += sortBy ? `&sort_by=${snakeCaseWord(sortBy)}` : '';
+    Object.keys(filteredData).forEach((key) => {
+      optionalUrlParams += `&${snakeCaseWord(key)}=${filteredData[key]}`;
+    });
+    return SubsidyApiService.apiClient().get(`${subsidiesURL}?page=${paginatedURL}${optionalUrlParams}`);
```
adamstankiewicz: [clarification] Is the `paginatedURL` variable a bit misleading? As it appears here, its value should be an integer (e.g., page **3**), but the variable name implies a string url?

src/data/services/SubsidyApiService.js
```diff
@@ -1,9 +1,27 @@
...
+    pageSize,
+    sortBy,
+    filteredData,
+  }) => {
+    const subsidiesURL = `${getConfig().SUBSIDY_BASE_URL}/api/v1/subsidies/`;
+    let optionalUrlParams = '';
+
+    optionalUrlParams += pageSize ? `&page_size=${pageSize}` : '';
+    optionalUrlParams += sortBy ? `&sort_by=${snakeCaseWord(sortBy)}` : '';
+    Object.keys(filteredData).forEach((key) => {
```
adamstankiewicz: [clarification] Is it true that the only key that should exist on this object here be `enterpriseCustomerUuid`? If that's the case, I wonder if it's worth passing in an explicit `enterpriseCustomerUuid` function argument instead of `filteredData` which implies multiple data attributes? (I may be misinterpreting; if so, please ignore 😄).

src/Configuration/Provisioning/data/hooks.js
```diff
@@ -2,24 +2,55 @@ import { useCallback } from 'react';
...
-    const fetchedData = camelCaseObject(sampleDataTableData(count, USES_LOCAL_TEST_DATA));
-    const normalizedData = normalizeSubsidyDataTableData({ fetchedData, actionIcon, redirectURL });
+
+  const hydrateEnterpriseSubsidies = useCallback(async ({ pageIndex, sortBy, filterBy }) => {
+    // Retrieve Basic List
+    const customerData = await LmsApiService.fetchEnterpriseCustomersBasicList();
+    const fetchedCustomerData = camelCaseObject(customerData.data);
+
+    // Filter by enterprise customer uuid for the enterprise customer uuid
+    const filteredData = filterByEnterpriseCustomerName({ fetchedCustomerData, filterBy });
```
adamstankiewicz, line 26: [clarification/curious] Are you needing to make this API call and do subsequent filtering in order to find the UUID associated (if any) with the enterprise name given in the filter input? I'm wondering if you've explored any possibilities to get `enterprise-subsidies` to be aware of enterprise customer names such that this API call can be removed in favor of the API having knowledge of the enterprise customer names, saving an API call on the client per table query. It may be worth at least scoping a couple ideas to see if the juice is worth the squeeze. 

For example, `enterprise-catalog`'s `EnterpriseCatalog` model has an attribute for enterprise customer name that we store along with the catalog UUID (example: http://localhost:18160/admin/catalog/enterprisecatalog/). I wonder if we did something similar, whether it could simplify this implementation and have other benefits beyond this one use case (e.g., all subsidy-related APIs could filter by enterprise name, all Django Admin screens could display/search by enterprise name). A potential challenge might be how to avoid stale data (e.g., if a enterprise name updates in the LMS, would `enterprise-subsidy` be aware of the change?).

Given this is an internal support tool, the extra performance hit to make the API call to retrieve the mapping of enterprise names -> enterprise UUIDs likely isn't too much of a concern. That said, having to do this seemingly extra API call for something learner/admin facing, maybe be too expensive performance wise IMHO.

src/Configuration/Provisioning/data/utils.js
```diff
@@ -244,24 +244,27 @@ export function getCamelCasedConfigAttribute(attribute) {
...
  * @returns - The normalized data to be displayed in the table
  */
-export function normalizeSubsidyDataTableData({ fetchedData, actionIcon, redirectURL }) {
-  if (fetchedData.length === 0) {
-    return [];
+export function normalizeSubsidyDataTableData({ fetchedSubsidyData, fetchedCustomerData }) {
+  if (fetchedSubsidyData.count === 0) {
+    return {
+      ...fetchedSubsidyData,
+      results: [],
```
adamstankiewicz: [clarification/question] If `fetchedSubsidyData.count === 0`, wouldn't its `results` already be an empty array? I guess the question is what is value of the default `fetchedSubsidyData.results` when `fetchedSubsidyData.count === 0`?

src/Configuration/Provisioning/data/utils.js
```diff
@@ -441,3 +444,69 @@ export function generatePolicyName(formData, index) {
...
+
+// Start of Datatable functions
+
+/**
+ * Takes a date string and returns a date string in the format of MM-DD-YYYY
+ * @param {String} date - The date string to be transformed
+ * @returns - Returns a date string in the format of MM-DD-YYYY
+ */
+export function transformDatatableDate(date) {
+  if (date) {
```
adamstankiewicz: nit: might recommend returning the `null` case first with `if (!date)` and then handling the happy path at the end. usually results in a bit cleaner code if/when the function grows.

src/utils/index.js
```diff
@@ -43,7 +42,19 @@ export const isValidCourseID = (value) => Boolean(value && value.match(COURSE_ID
...
+  if (value?.length <= 2) {
+    return Boolean(value && value.match(/^0{1,2}$/));
+  }
+  if (value?.length === 3) {
+    return Boolean(value && value.match(/^0{2}k$/));
+  }
+  if (value?.length > 3 || value?.length < 19) {
+    return Boolean(value && value.match(/^0{2}k([0-9A-Za-z]{1,15})$/));
+  }
+  return false;
```
adamstankiewicz: nit: this might be slightly simplified by failing early with `if (!value) { return false; }` and then the `?.` in the other conditionals wouldn't be necessary.

src/utils/index.js
```diff
@@ -62,6 +73,16 @@ export function titleCase(str) {
   return str.toLowerCase().replace(/_/g, ' ').replace(/\b(\w)/g, s => s.toUpperCase());
 }
 
+/**
+ * Convert a string containing camelCase into snake_case.
+ * @param {String} word - word to be converted to snake_case
+ * @returns - snake_case word
+ */
+export function snakeCaseWord(word) {
```
adamstankiewicz: [curious] I'm wondering if there's a chance you could use `snakeCaseObject` from `@edx/frontend-platform` in some way as opposed to defining a custom `snakeCase` function? That said, the functions do serve slightly different use cases (i.e., transforming a string vs. an object). Just a curiosity type question :)

src/Configuration/Provisioning/data/utils.js
```diff
@@ -244,24 +244,27 @@ export function getCamelCasedConfigAttribute(attribute) {
...
+      ...rest
     } = item;
-    const redirectUrl = () => redirectURL(uuid);
     return {
       ...rest,
-      uuid,
-      activeDatetime: new Date(activeDatetime).toLocaleDateString().replace(/\//g, '-'),
-      expirationDatetime: new Date(expirationDatetime).toLocaleDateString().replace(/\//g, '-'),
-      actions: actionIcon(redirectUrl),
+      enterpriseCustomerName: fetchedCustomerData.find(({ id }) => id === enterpriseCustomerUuid)?.name ?? '',
```
adamstankiewicz, line 258: [aside] another place that would be simplified if the API returned the enterprise customer name itself.

src/Configuration/Provisioning/data/utils.js
```diff
@@ -441,3 +444,69 @@ export function generatePolicyName(formData, index) {
...
+  return filterObj;
+}
+
+/**
+ * Takes the sort object from the datatable and returns a string that can be used to sort the API response
+ * @param {Object} sortBy - The sort object from the datatable
+ * @returns - Returns a string that can be used to sort the API response
+ */
+export function sortDatatableData({ sortBy }) {
+  if (sortBy[0]?.id) {
```
adamstankiewicz: nit: could likely lose a layer of conditional nesting, if failing early. E.g.,

```js
export function sortDatatableData({ sortBy }) {
  if (!sortBy?.[0]?.id) { return null; }
  if (sortBy[0].id === 'isActive') { return ... }
  return sortBy[0].desc ? `-${sortBy[0].id}` : sortBy[0].id;
}
```

adamstankiewicz: question: will there always be at most 1 element in `sortBy`?

nit: is it worth creating a variable for `sortBy[0]` so it doesn't need to be re-accessed in the below logic (may lead to more terse code 🤷‍♂️)

src/Configuration/Provisioning/data/hooks.js
adamstankiewicz: Follow-up from Slack: we'll opt to stick with the currently implemented pattern for now, due to the enterprise-subsidy service not being aware of the enterprise name right now. That said, having additional metadata about the enterprise customer across our services (i.e., data replication) seems to be a recurring need and it may be fruitful to think through a re-usable pattern for such data replication of enterprise customer metadata across services as a distinct chunk of work. This avoids introducing any scope creep for the current work in this PR, which is an internal support tool that can take the performance hit of an extra API call per `DataTable` fetch.

src/data/services/SubsidyApiService.js
brobro10000: So given multiple filters are applied (filter by plan title, filter by customer name => enterpriseCustomerUuid) it'll snake_case the key and add the value to construct the query parameter. Given the revamped implementation using `URLSearchParams` this entire section has been refactored away.

Screenshot 2023-07-11 at 10 24 30 AM

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormCustomer.jsx
brobro10000: Did not realize this was put in as a comment review, please ignore

src/Configuration/Provisioning/data/hooks.js
brobro10000: I can write up a new pair of tickets, one for enterprise-subsidy to update the API to support enterpriseCustomerName, the other for FE-app-support-tools to refactor the implementation to take into account the new API change.

src/Configuration/Provisioning/data/utils.js
brobro10000: Given the comment above, defering on this change, though I agree wholeheartedly

package.json
```diff
@@ -41,8 +41,8 @@
     "babel-polyfill": "6.26.0",
     "classnames": "2.2.6",
     "lodash.debounce": "4.0.8",
+    "lodash.snakecase": "^4.1.1",
```
adamstankiewicz: nit: pin the dependency for consistency with others

src/Configuration/Provisioning/DashboardDataTable/DashboardDataTable.jsx
```diff
@@ -0,0 +1,109 @@
...
+import DashboardTableBadges from './DashboardTableBadges';
+import { filterDatatableData, sortDataTableData, transformDatatableDate } from '../data/utils';
+
+const DashboardDataTable = () => {
+  const { enterpriseSubsidies } = useContextSelector(DashboardContext, v => v[0]);
+  const { hydrateEnterpriseSubsidies } = useDashboardContext();
+  const [isLoading, setIsLoading] = useState(true);
+
+  // Implementation due to filterText value displaying accessor value customerName as opposed to Customer Name
+  const filterStatus = (rest) => <DataTable.FilterStatus showFilteredFields={false} {...rest} />;
```
adamstankiewicz: As this variable is returning JSX, its a functional React component and should follow the convention of capitalized words (i.e., `FilterStatus`).

However, when doing this, it introduces a (valid) ESLint error around creating "unstable nested components": https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-unstable-nested-components.md

> Creating components inside components (nested components) will cause React to throw away the state of those nested components on each re-render of their parent.

Would recommend defining this function component outside of `DashboardDataTable` to appease the

src/Configuration/Provisioning/DashboardDataTable/DashboardDataTable.jsx
```diff
@@ -0,0 +1,109 @@
...
+      setIsLoading(false);
+    }
+  }, [hydrateEnterpriseSubsidies]);
+
+  const debouncedFetchData = useMemo(() => debounce(
+    fetchData,
+    300,
+    {
+      leading: false,
+    },
```
adamstankiewicz: [curious] I'm wondering the rationale for including `leading: false` here? It appears the default value for `debounce` is already `leading: false` ([source](https://github.com/lodash/lodash/blob/ddfd9b11a0126db2302cb70ec9973b66baec0975/lodash.js#L10312))?

src/Configuration/Provisioning/DashboardDataTable/DashboardDataTable.jsx
```diff
@@ -0,0 +1,109 @@
...
+          {
+            Header: 'Customer name',
+            accessor: 'enterpriseCustomerName',
+            disableSortBy: true,
+          },
+          {
+            Header: 'Start date',
+            accessor: 'activeDatetime',
+            disableFilters: true,
+            Cell: ({ row }) => transformDatatableDate(row.values.activeDatetime),
```
adamstankiewicz, line 86: nit: just a caution that if/when these functions were to return JSX, this line and line 94 below would be going against the `no-unstable-nested-components` ESLint rule. Not suggesting you need to change anything here, but just wanted to inform.

adamstankiewicz, line 86: nitty nit: `transformDataTableDate`

src/Configuration/Provisioning/DashboardDataTable/DashboardTableActions.jsx
```diff
@@ -0,0 +1,64 @@
...
+} from '@edx/paragon/icons';
+import PropTypes from 'prop-types';
+import ROUTES from '../../../data/constants/routes';
+
+const DashboardTableActions = ({ row }) => {
+  const rowUuid = row.values.uuid;
+  const { DJANGO_ADMIN_SUBSIDY_BASE_URL } = getConfig();
+  const history = useHistory();
+  const { HOME } = ROUTES.CONFIGURATION.SUB_DIRECTORY.PROVISIONING;
+  const enabledActionArray = [{
```
adamstankiewicz: [suggestion] As opposed to adding some complexity with the `enabled` property for each action and then filtering, could you do something like instead?:

```jsx
const actions = [
  <Hyperlink
    key="django-icon"
    destination={`${DJANGO_ADMIN_SUBSIDY_BASE_URL}/admin/subsidy/subsidy/?uuid=${rowUuid}`}
    target="_blank"
    showLaunchIcon={false}
    data-testid="django-admin-link"
  >
    <IconButton
      size="sm"
      src={DjangoShort}
      iconAs={Icon}
      alt="Django Admin Icon Button"
      data-testid={`Django-Admin-Page-${rowUuid}`}
    />
  </Hyperlink>,
];

if (getConfig().FEATURE_CONFIGURATION_EDIT_ENTERPRISE_PROVISION) {
  // conditionally prepend edit action
  actions.unshift(
    <IconButton
      key="edit-icon"
      size="sm"
      src={EditOutline}
      iconAs={Icon}
      onClick={() => history.push(`${HOME}/${rowUuid}/edit`)}
      alt="Edit Subsidy Icon Button"
      data-testid={`Edit-${rowUuid}`}
    />
  );
}
```

src/Configuration/Provisioning/DashboardDataTable/tests/DashboardTableActions.test.jsx
```diff
@@ -0,0 +1,57 @@
...
+    const row = {
+      values: {
+        uuid: 'Pikachu',
+      },
+    };
+    renderWithRouter(<DashboardTableActionsWrapper row={row} />);
+    expect(screen.getByTestId(`Edit-${row.values.uuid}`)).toBeTruthy();
+    expect(screen.getByTestId(`Django-Admin-Page-${row.values.uuid}`)).toBeTruthy();
+  });
+  it('useHistory is called on click, Edit', async () => {
```
adamstankiewicz: nit: might rename test to be less about the implementation details and more about the user behavior, e.g., `it('links to subsidy edit route').

src/Configuration/Provisioning/DashboardDataTable/tests/DashboardTableActions.test.jsx
```diff
@@ -0,0 +1,57 @@
...
+    const row = {
+      values: {
+        uuid: 'Pikachu',
+      },
+    };
+    renderWithRouter(<DashboardTableActionsWrapper row={row} />);
+    const editButton = screen.getByTestId(`Edit-${row.values.uuid}`);
+    fireEvent.click(editButton);
+    expect(useHistoryPush).toHaveBeenCalledWith(`${HOME}/${row.values.uuid}/edit`);
```
adamstankiewicz: nit: might also be worth including a `.toHaveBeenCalledTimes(1)` to assert the function is only called the expected number of times (1, in this case I'm assuming).

src/Configuration/Provisioning/data/utils.js
```diff
@@ -441,3 +441,70 @@ export function generatePolicyName(formData, index) {
...
+  }
+  return new Date(date).toLocaleDateString().replace(/\//g, '-');
+}
+
+/**
+ * Takes the filter object from the datatable and returns an object that can be used to filter the API response
+ * @param {Object} filters - The filter object from the datatable
+ * @returns - Returns an object that can be used to filter the API response
+ */
+export function filterDatatableData({ filters }) {
```
adamstankiewicz: nit: `filterDataTableData` and `transformDataTableDate`

src/Configuration/Provisioning/data/utils.js
```diff
@@ -441,3 +441,70 @@ export function generatePolicyName(formData, index) {
...
+ */
+export function transformDatatableDate(date) {
+  if (!date) {
+    return null;
+  }
+  return new Date(date).toLocaleDateString().replace(/\//g, '-');
+}
+
+/**
+ * Takes the filter object from the datatable and returns an object that can be used to filter the API response
```
adamstankiewicz: nit: it might be worth clarifying this comment a bit more. For example, are these filters being passed to the API request or is this filtering the API response returned by the request? I believe it's the former but the comment as written could be considered slightly misleading given its referring to filtering the _response_.

src/Configuration/Provisioning/DashboardDataTable/DashboardDataTable.jsx
brobro10000: Good point, will leave open for reference 👍🏽

brobro10000: Same as above, 👍🏽

src/Configuration/Provisioning/DashboardDataTable/DashboardDataTable.jsx
```diff
@@ -0,0 +1,107 @@
...
+import { sortDataTableData, transformDataTableData, transformDatatableDate } from '../data/utils';
+
+const FilterStatus = (rest) => <DataTable.FilterStatus showFilteredFields={false} {...rest} />;
+
+const DashboardDataTable = () => {
+  const { enterpriseSubsidies } = useContextSelector(DashboardContext, v => v[0]);
+  const { hydrateEnterpriseSubsidies } = useDashboardContext();
+  const [isLoading, setIsLoading] = useState(true);
+
+  // Implementation due to filterText value displaying accessor value customerName as opposed to Customer Name
```
adamstankiewicz: [clarification] Is this comment dangling from the now-moved `FilterStatus` above?

src/Configuration/Provisioning/DashboardDataTable/tests/DashboardTableActions.test.jsx
```diff
@@ -0,0 +1,58 @@
+/* eslint-disable react/prop-types */
+import { renderWithRouter } from '@edx/frontend-enterprise-utils';
+import { fireEvent, screen } from '@testing-library/react';
+import { Router } from 'react-router-dom';
+import DashboardTableActions from '../DashboardTableActions';
+import ROUTES from '../../../../data/constants/routes';
+
+const { HOME } = ROUTES.CONFIGURATION.SUB_DIRECTORY.PROVISIONING;
+
+const useHistoryPush = jest.fn();
```
adamstankiewicz: nit: `mockHistoryPush` might be a slightly more descriptive variable name

adamstankiewicz (approved):
LGTM. 2 additional, non-blocking little nits but :shipit:

src/Configuration/Provisioning/DashboardDataTable/DashboardDataTable.jsx
brobro10000: lol yeah it is my bad

PR #346, opened 2023-06-22
fix: PR updates

adamstankiewicz:


src/Configuration/Provisioning/data/utils.js
```diff
@@ -258,7 +258,6 @@ export function normalizeSubsidyDataTableData({ fetchedSubsidyData, actionIcon,
       uuid,
       activeDatetime: new Date(activeDatetime).toLocaleDateString().replace(/\//g, '-'),
```
adamstankiewicz, line 259: [consideration] If this is a display-only transformation, this may be better suited in the `Cell` attribute within the `columns` for `DataTable`.

brobro10000 (approved):

PR #355, opened 2023-09-06
feat: implement cancel button and modal

katrinan029:
# Description
As part of enhancing the user experience in the provisioning tool for LC2, we need to implement the ability to cancel changes made in the edit form. 

# Solution
The "Cancel" button is added to the top and bottom of the form in edit view. Additionally, a confirmation modal appears if the user leaves the edit mode with unsaved edits, allowing them to confirm their decision before discarding the changes.

https://github.com/openedx/frontend-app-support-tools/assets/71999631/c812dbbe-91ef-40d2-bdfd-4a84c0d981e2

[JIRA](https://2u-internal.atlassian.net/browse/ENT-7506)

### DIFF
--- a/src/Configuration/Provisioning/ProvisioningForm/PolicyForm/ProvisioningFormCatalog.jsx
+++ b/src/Configuration/Provisioning/ProvisioningForm/PolicyForm/ProvisioningFormCatalog.jsx
@@ -11,7 +11,12 @@ import { ProvisioningContext } from '../../ProvisioningContext';
 
 // TODO: Replace URL for hyperlink to somewhere to display catalog content information
 const ProvisioningFormCatalog = ({ index }) => {
-  const { setCustomCatalog, setCatalogQueryCategory, setInvalidPolicyFields } = useProvisioningContext();
+  const {
+    setCustomCatalog,
+    setCatalogQueryCategory,
+    setInvalidPolicyFields,
+    setHasEdits,
+  } = useProvisioningContext();
   const { CATALOG } = PROVISIONING_PAGE_TEXT.FORM;
   const contextData = useContextSelector(ProvisioningContext, v => v[0]);
   const {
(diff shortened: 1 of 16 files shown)


src/Configuration/Provisioning/SubsidyEditView/SubsidyEditView.jsx
```diff
@@ -35,6 +35,19 @@ const SubsidyEditView = () => {
     });
   };
 
+  useEffect(() => {
+    const handleTabClose = (event) => {
+      event.preventDefault();
+      const copyEvent = event;
+      copyEvent.returnValue = true;
+      return copyEvent.returnValue;
```
marlonkeating, line 46: Do we need to make a copy of the event here?
```suggestion
      const copyEvent = event;
      copyEvent.returnValue = true;
      return copyEvent.returnValue;
```

katrinan029: Yup! this was to get around the linter rule `no-param-reassign when setting a property on a DOM object`

marlonkeating: Do we need to set this?  Documentation suggests that `returnValue` is true by default (and also deprecated).
https://developer.mozilla.org/en-US/docs/Web/API/Event/returnValue

src/Configuration/Provisioning/ProvisioningForm/PolicyForm/ProvisioningFormCatalog.jsx
```diff
@@ -42,6 +48,9 @@ const ProvisioningFormCatalog = ({ index }) => {
   const handleChange = (e) => {
     const newTabValue = e.target.value;
     const newCatalogQuery = e.target.dataset.catalogqueryid;
+    if (isEditMode && !hasEdits) {
+      setHasEdits(true);
+    }
```
marlonkeating, line 53: Looks like these lines aren't being covered by tests.

marlonkeating (changes requested):

marlonkeating (approved):

PR #365, opened 2023-11-27
feat: Adds display name attribute to subsidy access policy creation API

brobro10000:
Adds `display_name` to the subsidy access policy POST method and related data transformation functions.

### DIFF
--- a/src/Configuration/Provisioning/data/utils.js
+++ b/src/Configuration/Provisioning/data/utils.js
@@ -439,6 +439,7 @@ export function transformSubsidyData(formData) {
  * @returns {Promise<Object>} - Returns a promise that resolves to the response data from the API
  */
 export async function createPolicy({
+  displayName,
   description,
   enterpriseCustomerUuid,
   catalogUuid,
@@ -447,6 +448,7 @@ export async function createPolicy({
   spendLimit,
 }) {
   const data = LmsApiService.postSubsidyAccessPolicy(
+    displayName,
     description,
     enterpriseCustomerUuid,
     catalogUuid,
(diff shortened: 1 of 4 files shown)


src/Configuration/Provisioning/data/utils.js
```diff
@@ -503,6 +504,7 @@ export function transformPolicyData(formData, catalogCreationResponse, subsidyCr
     || subsidyCreationResponse.length === 0
   ) { return []; }
   const payloads = policies.map((policy, index) => ({
+    displayName: policy.accountName,
```
adamstankiewicz, line 507: [nit, question, non-blocking] What does **account** refer to here? I'm wondering if it might make sense to more closely align the property names to how the API returns these attributes? That said, _if_ we take any action here, definitely for a separate PR given unrelated variable names, etc. would also need to change.

For example, at first glance, it's not immediately obvious to me that `accountName` is referring to the display name of a budget; initially, I had read "account" to mean mean the enterprise customer, not the budget itself (e.g., is `policy.accountName` the enterprise customer name?).

adamstankiewicz (approved):

src/Configuration/Provisioning/data/utils.js
brobro10000: Agreed on the sentiment for the renaming of the object attributes. I know theres an upcoming PR that revamps some of the naming scheme to be more understandable, so it was intentional to keep this PR as simple as possible.

But I will definitely leave this comment open as a reference to when we do have to take on that future work. 👍🏽

PR #369, opened 2023-12-05
fix: 3 bugs in the provisioning tool

pwnage101:
1. We no longer double-multiply the subsidy balance by 100, leading to a higher subsidy starting balance than intended.
2. catalogUuid no longer is interpreted as invalid when a predefined query is selected.
3. On the edit view, the budget header now correctly reflects the catalog selection instead of always displaying "Budget".  Now it displays "Open Courses budget", "Unique/Curated budget" , etc. depending on the catalog selection.

ENT-7922

### DIFF
--- a/src/data/services/SubsidyApiService.js
+++ b/src/data/services/SubsidyApiService.js
@@ -25,6 +25,22 @@ class SubsidyApiService {
     return SubsidyApiService.apiClient().get(`${subsidiesURL}?uuid=${uuid}`);
   };
 
+  /**
+   * postSubsidy gets or creates a learner credit Subsidy (and corresponding ledger).
+   *
+   * @param {String} financialIdentifier - A reference to the object responsible for originating this subsidy, and the
+   * key on which existing subsidies are retrieved.
+   * @param {String} title
+   * @param {String} enterpriseUUID
+   * @param {String} startDate
+   * @param {String} endDate
+   * @param {Number} startingBalance - The initial balance of the new subsidy in USD Cents (integer).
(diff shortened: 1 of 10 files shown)


src/data/services/SubsidyApiService.js
```diff
@@ -25,6 +25,22 @@ class SubsidyApiService {
...
+  /**
+   * postSubsidy gets or creates a learner credit Subsidy (and corresponding ledger).
+   *
+   * @param {String} financialIdentifier - A reference to the object responsible for originating this subsidy, and the
+   * key on which existing subsidies are retrieved.
+   * @param {String} title
+   * @param {String} enterpriseUUID
+   * @param {String} startDate
+   * @param {String} endDate
+   * @param {Number} startingBalance - The initial balance of the new subsidy in USD Cents (integer).
```
pwnage101, line 37: Our component-level tests don't cover testing this layer, so they don't assert what's _really_ being sent in the POST request.  The least I can do is document the expected parameters.

brobro10000 (approved):
LGTM. I would just double check the price validation in stage for the `startingBalance` on the subsidy. 👍🏽  Great job.

pwnage101:
FYI @brobro10000 I did just push one more commit to fix the new "policy type" field component to be supported in the edit view.  It was pretty straightforward and mostly copy+pasting from another component to add that support.

pwnage101:
Also added a commit to display the policy type in the edit view (readonly).
image

FYI @katrinan029

PR #370, opened 2023-12-08
fix: can create new plans with test plan unchecked

pwnage101:
Somehow we let a regression slip through manual QA testing and unit testing which prevented non-test plans from being created.  The provisioning tool would POST to create a subsidy, but failed to supply a `default_internal_only` request parameter, which is required.  This change adds a default value for formData.internalOnly on context initialization to prevent the field from ever being empty.

ENT-8107

testing done
===
1. create a new plan without checking the Test Plan checkbox.
2. create a new plan, checking the Test Plan checkbox.
3. edit an existing plan, unchecking the Test Plan checkbox and saving.
4. edit an existing plan, checking the Test Plan checkbox and saving.

### DIFF
--- a/src/Configuration/Provisioning/ProvisioningContext.jsx
+++ b/src/Configuration/Provisioning/ProvisioningContext.jsx
@@ -20,6 +20,8 @@ const ProvisioningContextProvider = ({ children }) => {
       // `policies` is a list of all policies/budgets inputs for this plan. After form completion, length should be 1 if
       // multipleFunds = false, and 2 and multipleFunds = true.
       policies: [],
+      // `internalOnly` backs the "Test plan" checkbox.  A true value means the plan will not be customer facing.
+      internalOnly: false,
     },
     // `showInvalidField` indicates when form fields are valid/invalid, possibly triggering invalid attributes to be set
     // on various form field containers.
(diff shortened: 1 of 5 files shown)


src/Configuration/Provisioning/ProvisioningContext.jsx
```diff
@@ -20,6 +20,8 @@ const ProvisioningContextProvider = ({ children }) => {
       // `policies` is a list of all policies/budgets inputs for this plan. After form completion, length should be 1 if
       // multipleFunds = false, and 2 and multipleFunds = true.
       policies: [],
+      // `internalOnly` backs the "Test plan" checkbox.  A true value means the plan will not be customer facing.
+      internalOnly: false,
```
pwnage101, line 24: This line fixes the bug.

src/Configuration/Provisioning/ProvisioningForm/ProvisioningFormInternalOnly.jsx
```diff
@@ -11,11 +11,9 @@ const ProvisioningFormInternalOnly = () => {
...
   const [formData, isEditMode, hasEdits] = selectProvisioningContext('formData', 'isEditMode', 'hasEdits');
 
-  let submittedFormInternalOnly;
-  if (isEditMode) {
-    submittedFormInternalOnly = formData.internalOnly;
-  }
-  const [value, setValue] = useState(submittedFormInternalOnly || false);
+  // formData.internalOnly is always defined, whether it's hydrated from an API call (isEditMode = true) or not
+  // (isEditMode = false) in which case ProvisioningContextProvider supplies a default.
+  const [value, setValue] = useState(formData.internalOnly || false);
```
pwnage101, line 16: Since we can assume that formData.internalOnly is always defined, we can kill the if-condition.

brobro10000 (approved):
LGTM 👍🏽  Great job

PR #373, opened 2024-01-05
ENT-7820 Learner Credit table sorting

jono-booth:
There were a couple issues with the `sortDataTableData` method.

1. It wasn't returning the result of the ternary, so needed to set the result to a var and return the var
2. `lodash.snakecase` was stripping out the `-` which is used for sorting in reverse. Moved the `snakecase` call before the `-` was prepended.

Added some test for asc and desc sorting.

### DIFF
--- a/src/Configuration/Provisioning/DashboardDataTable/tests/DashboardDataTable.test.jsx
+++ b/src/Configuration/Provisioning/DashboardDataTable/tests/DashboardDataTable.test.jsx
@@ -1,21 +1,21 @@
 /* eslint-disable react/prop-types */
 import { renderWithRouter } from '@edx/frontend-enterprise-utils';
 import { screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import { camelCaseObject } from '@edx/frontend-platform';
 import { DashboardContext, initialStateValue } from '../../../testData/Dashboard';
 import DashboardDataTable from '../DashboardDataTable';
 import { sampleDataTableData } from '../../../testData/constants';
+import SubsidyApiService from '../../../../data/services/SubsidyApiService';
 
 // Mock the debounce function
 jest.mock('lodash.debounce', () => jest.fn((fn) => fn));
 
 // Mock the subsidy list
 const mockGetAllSubsidiesData = sampleDataTableData(10);
-jest.mock('../../../../data/services/SubsidyApiService', () => ({
-  getAllSubsidies: () => Promise.resolve({
-    data: mockGetAllSubsidiesData,
-  }),
-}));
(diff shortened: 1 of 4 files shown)


src/Configuration/Provisioning/DashboardDataTable/tests/DashboardDataTable.test.jsx
```diff
@@ -47,3 +46,24 @@ describe('DashboardDatatable', () => {
     await waitFor(() => expect(screen.getByText('Enterprise Customer 1')).toBeTruthy());
   });
 });
+
+describe('DashboardDatatable SortBy', ()=> {
+  it('sorts the data asc and desc', async ()=> {
+    renderWithRouter(<DashboardDatatableWrapper />);
+    const tableHeader = screen.getByText('Start date');
+
+    fireEvent.click(tableHeader);
```
brobro10000: Instead of `fireEvent` the repo is leaning towards `userEvent` in testing.  See https://github.com/openedx/frontend-app-support-tools/blob/f00e83fd68ebd8001526ba7d2cf9a3bf9d4e2bcc/src/Configuration/Provisioning/ProvisioningForm/PolicyForm/tests/ProvisioningFormPolicyContainer.test.jsx#L69-L90  for an example of its usage

```js
import userEvent from '@testing-library/user-event';

const tableHeader = screen.getByText('Start date');

    userEvent.click(tableHeader);

```

src/Configuration/Provisioning/data/utils.js
```diff
@@ -650,10 +651,13 @@ export function sortDataTableData({ sortBy }) {
   if (!sortByObject) {
     return null;
   }
-  if (sortByObject.id === 'isActive') {
-    return sortByObject.desc ? '-expirationDatetime' : 'expirationDatetime';
+  let sortString = sortByObject.id;
+  if ( sortString === 'isActive') {
```
brobro10000: Noticed some linting errors on this PR. Running the script `npm run lint` should point out where those linting errors exists and how to correct them. 

Screenshot 2024-01-05 at 11 41 32 AM

brobro10000 (commented):
Running the PR locally, it is sorting as expected 👍🏽  but noticed some failing tests. Will run the CI on this for now while it gets corrected.

jono-booth:
> Running the PR locally, it is sorting as expected 👍🏽 but noticed some failing tests. Will run the CI on this for now while it gets corrected.

Thanks for the feedback. Sorted the linting issues out and using `userEvent` instead of `fireEvent` in the test.

brobro10000 (approved):
LGTM 👍🏽  Great job

PR #379, opened 2024-03-27
Add comment to course reset

hajorg:
### Summary
This PR allows for comment to be added when resetting a learner's course
See: https://2u-internal.atlassian.net/browse/AU-1935

Screenshot 2024-03-27 at 14 27 03
Screenshot 2024-03-27 at 14 28 41

### DIFF
--- a/src/users/CourseReset.jsx
+++ b/src/users/CourseReset.jsx
@@ -1,5 +1,5 @@
 import {
-  Alert, AlertModal, Button, useToggle, ActionRow,
+  Alert, AlertModal, Button, useToggle, ActionRow, Form,
 } from '@edx/paragon';
 import PropTypes from 'prop-types';
 import React, { useCallback, useEffect, useState } from 'react';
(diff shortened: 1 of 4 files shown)


jansenk (approved):
LGTM

src/users/CourseReset.jsx
```diff
@@ -17,8 +17,20 @@ function CourseReset({ username, intl }) {
   const [courseResetData, setCourseResetData] = useState([]);
   const [error, setError] = useState('');
   const [isOpen, open, close] = useToggle(false);
+  const [comment, setComment] = useState();
```
AfaqShuaib09:
```suggestion
const [comment, setComment] = useState('');
```

src/users/CourseReset.jsx
```diff
@@ -17,8 +17,20 @@ function CourseReset({ username, intl }) {
...
+  const [commentError, setCommentError] = useState('');
   const POLLING_INTERVAL = 10000;
 
+  const handleCommentChange = (e) => {
+    const text = e.target.value;
+    setComment(text);
+    if (text.length > 255) {
```
AfaqShuaib09: would be nice if we store this 255 in a variable
```javascript
MAX_COMMENT_LENGTH = 255
```
so we can use this in comment message error too

AfaqShuaib09 (approved):
Just a couple of suggestions otherwise looks good to me 👍🏻

PR #381, opened 2024-04-05
fix: make some of this app work in tutor

kdmccormick:
sandbox at: https://app.pr-381-98a602.sandboxes.opencraft.hosting/support-tools/learner_information/?username=openedx

### Description

This makes most of the support-tools pages render when running in Tutor. The main problem was that `getConfig()` was being run at init-time instead of runtime, so the MFE was using values from the .env files instead of from the MFE Config API.

### Running in Tutor

This just requires enabling tutor-mfe plus a simple custom plugin:
```python
from tutor.hooks import Filters
from tutormfe.hooks import MFE_APPS

@MFE_APPS.add()
def _add_support_tools(apps: dict) -> dict:
    return {
        **apps,
        "support-tools": {
            "repository": "https://github.com/openedx/frontend-app-support-tools",
            "port": 2007,
            "version": "master",
        },
    }
```

### Open issues

* Several features don't work because the feature they depend on is disabled in LMS/CMS, such as feature-based-enrollment.
* Several pages don't work because their routes assume they are running on the root (`/`) instead of a subpath (`/support-tools`). We have fixed this problem in several MFEs and it could be fixed here too.
* Several features are disabled because they depend on microservices: credentials, discovery, ecommerce, enterprise-catalog, license-manager.
* Some links are hard-coded to 2U accounts so they will never work outside of edx.org.

### Status

This was mostly for my own reference. I don't plan on trying to merge this unless it becomes clear that we want to keep this MFE in the openedx org.

### Some screenshots of pages that do work

st1
st2
st3


open-craft-grove:
Sandbox deployment successful 🚀
🎓 [LMS](https://pr-381-98a602.sandboxes.opencraft.hosting)
📝 [Studio](https://studio.pr-381-98a602.sandboxes.opencraft.hosting)
ℹ️ [Grove Config](https://axim-sandbox-build-logs-4626178240.nyc3.digitaloceanspaces.com/pr-381-98a602-51564656-6560824899-tutor-config.yml), [Tutor Config](https://axim-sandbox-build-logs-4626178240.nyc3.digitaloceanspaces.com/pr-381-98a602-51564656-6560824899-grove-config.yml), [Tutor Requirements](https://axim-sandbox-build-logs-4626178240.nyc3.digitaloceanspaces.com/pr-381-98a602-51564656-6560824899-tutor-requirements.txt)

PR #400, opened 2024-08-15
chore: remove request dev dependency

brobro10000:
Removes the `request` package from dev dependency which should resolve the following dependabot vulnerability alerts:
- https://github.com/openedx/frontend-app-support-tools/security/dependabot/82
- https://github.com/openedx/frontend-app-support-tools/security/dependabot/86


adamstankiewicz (approved):
LGTM; this one was a `tough-cookie` 🍪 😜

PR #414, opened 2024-09-09
fix: add integration count number

kiram15:
Adding a count of the number of integrations associated with that customer 
[Jira ticket ](https://2u-internal.atlassian.net/browse/ENT-9448)

Screenshot 2024-09-09 at 9 35 42 AM


katrinan029 (approved):
LGTM!

src/Configuration/Customers/CustomerDetailView/CustomerIntegrations.jsx
```diff
@@ -11,12 +11,16 @@ const CustomerIntegrations = ({
   const { ADMIN_PORTAL_BASE_URL } = getConfig();
   const ssoDateText = ({ sso }) => (`Created ${formatDate(sso?.created)} • Last modified ${formatDate(sso?.modifed)}`);
   const configDateText = ({ config }) => (`Created ${formatDate(config?.created)} • Last modified ${formatDate(config?.lastModifiedAt)}`);
+  let integrationCount = activeSSO?.length + activeIntegrations?.length;
```
marlonkeating: Is there any chance that `activeSSO` or `activeIntegrations` will be undefined?  If so, we're going to get either `undefined` or `NaN` out of this expression, which will mess up all future operations on this value.

marlonkeating (approved):

PR #418, opened 2024-09-12
feat: Removed IDV Status details call

rijuma:
# Description

After this change on `edx-name-affirmation` (https://github.com/edx/edx-name-affirmation/pull/215), the status for the IDV attempt status was added to the payload for the Verified Name History endpoint. This change removes the need for using the account verification detail to get the attempt status.

I've also tweaked the tooltip to make it a bit nicer on the UI and formatted properly for human readability.

# Screenshots
Screenshot 2024-09-13 at 12 25 34

Screenshot 2024-09-13 at 12 25 46


Ali-D-Akbar (approved):

src/users/VerifiedName.jsx
```diff
@@ -10,13 +10,68 @@ import {
...
+    accessor: 'proctoringAttemptId',
+  },
+  {
+    Header: 'Created At',
+    accessor: 'createdAt',
+  },
+];
+
+// Human readable formatter for the status. Possible status list on:
+// https://github.com/edx/edx-solutions-edx-platform/blob/0ebc69f86548a44b7947decfe308032028721907/lms/djangoapps/verify_student/models.py#L104
```
michaelroytman: Nit: The `edx-solutions-platform` is actually an out-of-date fork of [edx-platform](https://github.com/openedx/edx-platform). Could you include a link to these lines in the `openedx` version of `edx-platform`, please?

src/users/VerifiedName.jsx
```diff
@@ -10,13 +10,68 @@ import {
...
+const idvStatusFormat = status => {
```
michaelroytman: I'm a little on the fence about this formatting. On the one hand, I think it makes it more readable. On the other hand, the status of the VerifiedName isn't formatted and is the plain database status, but this one is formatted, so it's inconsistent. I feel like it may be more useful to just expose the plain database status here so that staff can correlate the values here with what's in the database without needing to translate. What do you think?

michaelroytman: I'm also apprehensive because this is a change to the existing behavior, and I haven't reviewed that with product or Axim. I'd prefer we keep the formatting as it. But perhaps this is a topic for stand up?

rijuma: Yes, sure. I've made this change thinking about making it more user friendly, but it is a change in behavior. I'll just remove it, no problem.

src/users/VerifiedName.jsx
rijuma: Removed formatting functionality.

src/users/VerifiedName.jsx
rijuma: Removed formatting functionality.

michaelroytman (approved):
Looks great! Thank you for addressing my feedback and making the UI clearer!

zawan-ila (approved):

PR #426, opened 2024-09-25
feat: add platform verification attempt column to verified name history

alangsto:
## [COSMO-496](https://2u-internal.atlassian.net/browse/COSMO-496)

This is part of the work for https://github.com/openedx/platform-roadmap/issues/367.

The Verified Name History view surfaces information to support regarding how a name was verified. Because PSI IDV verification is deprecated, and we are integrating with a new vendor (Persona), we should modify the Verified Name History view to account for these changes. 

The PSI IDV Verification information has been renamed to include "Legacy", and any attempts created through Persona will be under the "IDV Attempt ID" column. 

This is the new view:
Screenshot 2024-09-25 at 2 52 53 PM


schenedx (approved):
👍

Ali-D-Akbar (approved):

PR #443, opened 2025-01-21
fix: adding typing delay to associated users table

kiram15:
[Jira Ticket ](https://2u-internal.atlassian.net/browse/ENT-9554)

https://github.com/user-attachments/assets/79f98e9d-e0b0-451c-8704-80dab7333903

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,7 @@
     "@testing-library/dom": "^9.3.4",
     "@testing-library/jest-dom": "^5.17.0",
     "@testing-library/react": "12.1.4",
+    "@testing-library/react-hooks": "^8.0.1",
     "@testing-library/user-event": "^14.5.2",
     "@types/react-table": "^7.7.2",
     "@wojtekmaj/enzyme-adapter-react-17": "^0.8.0",
--- a/src/Configuration/Customers/data/hooks/useCustomerUsersTableData.js
+++ b/src/Configuration/Customers/data/hooks/useCustomerUsersTableData.js
@@ -19,7 +19,6 @@ const useCustomerUsersTableData = (enterpriseUuid) => {
       try {
         setIsLoading(true);
         const options = {};
-
         args.sortBy.filter((sort) => {
           const { id, desc } = sort;
           if (id === 'administrator') {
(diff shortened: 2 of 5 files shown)


src/Configuration/Customers/data/hooks/useCustomerUsersTableData.js
```diff
@@ -56,8 +56,13 @@ const useCustomerUsersTableData = (enterpriseUuid) => {
         setIsLoading(false);
       }
     };
+
     if (enterpriseUuid) {
-      fetch();
+      if (args.filters.length && args.filters[0].value.length > 2) {
```
marlonkeating: [question] Do we know that the typing filter will always be first in the list of filters?

src/Configuration/Customers/data/hooks/useCustomerUsersTableData.js
```diff
@@ -56,8 +56,13 @@ const useCustomerUsersTableData = (enterpriseUuid) => {
...
+        fetch();
+      } else if (!args.filters.length) {
+        fetch();
+      }
```
marlonkeating: nit: Can condense this logic
```suggestion
      if (!args.filters.length || args.filters[0].value.length > 2) {
        fetch();
      }
```

marlonkeating (changes requested):
Can we add a unit test to `EnterpriseCustomerUsersTable.test.jsx`?

src/Configuration/Customers/data/hooks/useCustomerUsersTableData.js
kiram15: Yes, its the only filter we have for this table

package.json
```diff
@@ -40,6 +40,7 @@
     "@fortawesome/free-solid-svg-icons": "5.15.1",
     "@fortawesome/react-fontawesome": "^0.1.14",
     "@openedx/paragon": "21.13.1",
+    "@testing-library/react-hooks": "^8.0.1",
```
marlonkeating: As this is only used in testing, can we put this in `devDependencies`?

marlonkeating (approved):

PR #452, opened 2025-03-24
feat: remove PII from urls

zawan-ila:
# [PROD-4319](https://2u-internal.atlassian.net/browse/PROD-4319)

This PR makes the following changes:-

1. Update the Learner Information Page to append lms_user_id in the query params even in the case when the search is through a username or email.
2. Update the Program Inspector page to append the edx_user_id in the query params instead of the edx_user, org_key and external_user_key params.

These changes are intended to remove PII like emails and usernames from the urls.

### DIFF
--- a/src/ProgramEnrollments/ProgramInspector/ProgramInspector.jsx
+++ b/src/ProgramEnrollments/ProgramInspector/ProgramInspector.jsx
@@ -49,7 +49,7 @@ export default function ProgramInspector() {
       if (newLink === location.search) {
         setClickEventCall(!clickEventCall);
       } else {
-        setQuery(newLink)
+        setQuery(newLink);
       }
     }
   };
@@ -68,36 +68,39 @@ export default function ProgramInspector() {
         setError(response.error);
         setActiveOrgKey(response.org_keys);
         setLearnerProgramEnrollment(response.learner_program_enrollments);
-        let name = response?.learner_program_enrollments?.user?.username;
+        const name = response?.learner_program_enrollments?.user?.username;
         return name;
-      }).then((username) => {
-        if (username) {
-          getUser(username).then(
-            res => navigate(`?edx_user_id=${res.id}`)
-          )
+      }).then((name) => {
+        if (name) {
+          getUser(name).then(
+            res => navigate(`?edx_user_id=${res.id}`),
+          );
         }
       }).catch(err => {
+        console.error(err);
         setError('An error occured while fetching user id');
-        navigate('/programs')
-      });;
+        navigate('/programs');
+      });
     }
   };
 
   useEffect(() => {
-    if (query)
+    if (query) {
       fetchInspectorData(query);
+    }
   }, [query]);
 
   useEffect(() => {
-    let userId;
-    if (userId = new URLSearchParams(location.search).get('edx_user_id')) {
+    const userId = new URLSearchParams(location.search).get('edx_user_id');
+    if (userId) {
       getUser(userId).then(res => {
         setUsername(res.username);
-        setQuery(`?edx_user=${res.username}&org_key=${activeOrgKey}&external_user_key=${externalUserKey}`)
+        setQuery(`?edx_user=${res.username}&org_key=${activeOrgKey}&external_user_key=${externalUserKey}`);
       }).catch(err => {
+        console.error(err);
         setError('An error occured while fetching user id');
-        navigate('/programs')
-      })
+        navigate('/programs');
+      });
     }
   }, []);
 
--- a/src/ProgramEnrollments/ProgramInspector/ProgramInspector.jsx
+++ b/src/ProgramEnrollments/ProgramInspector/ProgramInspector.jsx
@@ -75,7 +75,7 @@ export default function ProgramInspector() {
       })
         .catch(err => {
           console.error(err);
-          setError('Five errors occurred');
+          setError('An error occured while fetching user id');
           navigate('/programs');
         });
     }
--- a/src/ProgramEnrollments/ProgramInspector/ProgramInspector.jsx
+++ b/src/ProgramEnrollments/ProgramInspector/ProgramInspector.jsx
@@ -70,17 +70,14 @@ export default function ProgramInspector() {
         setLearnerProgramEnrollment(response.learner_program_enrollments);
         const name = response?.learner_program_enrollments?.user?.username;
         return name;
-      }).then((name) => {
-        if (name) {
-          getUser(name).then(
-            res => navigate(`?edx_user_id=${res.id}`),
-          );
-        }
-      }).catch(err => {
-        console.error(err);
-        setError('An error occured while fetching user id');
-        navigate('/programs');
-      });
+      }).then((name) => getUser(name)).then((res) => {
+        navigate(`?edx_user_id=${res.id}`);
+      })
+        .catch(err => {
+          console.error(err);
+          setError('Five errors occurred');
+          navigate('/programs');
+        });
     }
   };
 
--- a/src/ProgramEnrollments/ProgramInspector/ProgramInspector.jsx
+++ b/src/ProgramEnrollments/ProgramInspector/ProgramInspector.jsx
@@ -24,7 +24,6 @@ export default function ProgramInspector() {
   const [activeOrgKey, setActiveOrgKey] = useState(params.get('org_key'));
   const [orgKeyList, setOrgKeyList] = useState(undefined);
   const [externalUserKey, setExternalUserKey] = useState(params.get('external_user_key'));
-  const [clickEventCall, setClickEventCall] = useState(false);
 
   const [query, setQuery] = useState(null);
 
@@ -43,14 +42,10 @@ export default function ProgramInspector() {
       setSsoRecords([]);
       navigate('/programs');
     } else {
-      const newLink = `?edx_user=${
+      const newQuery = `?edx_user=${
         username || ''
       }&org_key=${activeOrgKey}&external_user_key=${externalUserKey || ''}`;
-      if (newLink === location.search) {
-        setClickEventCall(!clickEventCall);
-      } else {
-        setQuery(newLink);
-      }
+      setQuery(newQuery);
     }
   };
 
@@ -75,7 +70,7 @@ export default function ProgramInspector() {
       })
         .catch(err => {
           console.error(err);
-          setError('An error occured while fetching user id');
+          setError('An error occurred while fetching user id');
           navigate('/programs');
         });
     }
@@ -95,7 +90,7 @@ export default function ProgramInspector() {
         setQuery(`?edx_user=${res.username}&org_key=${activeOrgKey}&external_user_key=${externalUserKey}`);
       }).catch(err => {
         console.error(err);
-        setError('An error occured while fetching user id');
+        setError('An error occurred while fetching user id');
         navigate('/programs');
       });
     }
--- a/src/ProgramEnrollments/ProgramInspector/ProgramInspector.jsx
+++ b/src/ProgramEnrollments/ProgramInspector/ProgramInspector.jsx
@@ -3,7 +3,7 @@ import { useLocation, useNavigate } from 'react-router-dom';
 import {
   Alert, Col, Row, Button, Input,
 } from '@openedx/paragon';
-import { getSsoRecords } from '../../users/data/api';
+import { getSsoRecords, getUser } from '../../users/data/api';
 import EnrollmentDetails from './EnrollmentDetails';
 import SingleSignOnRecordCard from '../../users/SingleSignOnRecordCard';
 import {
@@ -26,6 +26,8 @@ export default function ProgramInspector() {
   const [externalUserKey, setExternalUserKey] = useState(params.get('external_user_key'));
   const [clickEventCall, setClickEventCall] = useState(false);
 
+  const [query, setQuery] = useState(null);
+
   const getOrgKeyList = () => (orgKeyList
     ? orgKeyList.map((data) => ({
       value: data,
@@ -41,13 +43,13 @@ export default function ProgramInspector() {
       setSsoRecords([]);
       navigate('/programs');
     } else {
-      const newLink = `/programs?edx_user=${
+      const newLink = `?edx_user=${
         username || ''
       }&org_key=${activeOrgKey}&external_user_key=${externalUserKey || ''}`;
-      if (newLink === location.pathname + location.search) {
+      if (newLink === location.search) {
         setClickEventCall(!clickEventCall);
       } else {
-        navigate(newLink);
+        setQuery(newLink)
       }
     }
   };
@@ -66,13 +68,38 @@ export default function ProgramInspector() {
         setError(response.error);
         setActiveOrgKey(response.org_keys);
         setLearnerProgramEnrollment(response.learner_program_enrollments);
-      });
+        let name = response?.learner_program_enrollments?.user?.username;
+        return name;
+      }).then((username) => {
+        if (username) {
+          getUser(username).then(
+            res => navigate(`?edx_user_id=${res.id}`)
+          )
+        }
+      }).catch(err => {
+        setError('An error occured while fetching user id');
+        navigate('/programs')
+      });;
     }
   };
 
   useEffect(() => {
-    fetchInspectorData(location.search);
-  }, [location.search, clickEventCall]);
+    if (query)
+      fetchInspectorData(query);
+  }, [query]);
+
+  useEffect(() => {
+    let userId;
+    if (userId = new URLSearchParams(location.search).get('edx_user_id')) {
+      getUser(userId).then(res => {
+        setUsername(res.username);
+        setQuery(`?edx_user=${res.username}&org_key=${activeOrgKey}&external_user_key=${externalUserKey}`)
+      }).catch(err => {
+        setError('An error occured while fetching user id');
+        navigate('/programs')
+      })
+    }
+  }, []);
 
   useEffect(() => {
     if (!orgKeyList) {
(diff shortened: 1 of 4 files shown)


src/users/UserPage.jsx
```diff
@@ -137,16 +131,7 @@ export default function UserPage() {
...
-  }, [params.get('username'), params.get('email'), params.get('lms_user_id')]);
-
-  // To change the url with appropriate query param if query param info is not present in URL
-  useLayoutEffect(() => {
-    if (userIdentifier
-      && location.pathname.indexOf(TAB_PATH_MAP[LEARNER_INFO_TAB]) !== -1
-      && !(params.get('email') || params.get('username') || params.get('lms_user_id'))) {
-      pushHistoryIfChanged(getUpdatedURL(userIdentifier));
-    }
-  });
```
DawoudSheraz, line 149: why is this removed?

src/users/UserPage.jsx
```diff
@@ -137,16 +131,7 @@ export default function UserPage() {
...
+  }, []);
```
DawoudSheraz, line 133: Won't we want lms_user_id to remain in place? I think this was added in case someone edits the param directly in URL and hits the URL. The new code only runs this on the first render.

zawan-ila: If the user edits param in URL and hits it, I'd expect that to cause a complete reload of the page. (leading to the "first" render). Personally, I was unable to see any reason for this useEffect and hence removed it. The updation of the url in `processSearchResult` seems to do the job fine. Maybe @Ali-D-Akbar can shed some light here, if he can rev his brain back to when he added it (3 years ago :-))

src/users/UserPage.jsx
zawan-ila: responded in comment above

src/ProgramEnrollments/ProgramInspector/ProgramInspector.jsx
```diff
@@ -66,13 +68,38 @@ export default function ProgramInspector() {
         setError(response.error);
         setActiveOrgKey(response.org_keys);
         setLearnerProgramEnrollment(response.learner_program_enrollments);
-      });
+        const name = response?.learner_program_enrollments?.user?.username;
+        return name;
+      }).then((name) => getUser(name)).then((res) => {
+        navigate(`?edx_user_id=${res.id}`);
```
DawoudSheraz, line 69: should we not have query in here as well alongside the user id?

src/ProgramEnrollments/ProgramInspector/ProgramInspector.test.jsx
```diff
@@ -38,23 +40,32 @@ describe('Program Inspector', () => {
...
+  let getUserMock;
 
   const data = {
     username: 'verified',
     orgKey: samlProvidersResponseValues[0],
     externalKey: 'testuser',
   };
 
+  const updateWrapperState = async (w) => {
+    await new Promise((res) => { setTimeout(() => res(), 0); });
```
DawoudSheraz: - What does res correspond to here? 
- Instead of w, please use complete word (wrapper) here.

src/ProgramEnrollments/ProgramInspector/ProgramInspector.test.jsx
```diff
@@ -68,24 +79,26 @@ describe('Program Inspector', () => {
...
-    expect(externalKeyInput.prop('defaultValue')).toEqual('');
+    expect(usernameInput.prop('defaultValue')).toEqual(undefined);
+    expect(externalKeyInput.prop('defaultValue')).toEqual(undefined);
   });
 
   it('render when username', async () => {
     apiMock = jest
       .spyOn(api, 'getProgramEnrollmentsInspector')
-      .mockImplementationOnce(() => Promise.resolve(programInspectorSuccessResponse));
+      .mockImplementation(() => Promise.resolve(programInspectorSuccessResponse));
```
DawoudSheraz, line 94: Why is this no longer mocked once? The flow is adding a call to get user info, this should not have been affected.

src/ProgramEnrollments/ProgramInspector/ProgramInspector.test.jsx
```diff
@@ -68,24 +79,26 @@ describe('Program Inspector', () => {
...
       .spyOn(api, 'getProgramEnrollmentsInspector')
-      .mockImplementationOnce(() => Promise.resolve(programInspectorErrorResponse));
+      .mockImplementation(() => Promise.resolve(programInspectorErrorResponse));
 
     const usernameInput = wrapper.find("input[name='username']");
     const externalKeyInput = wrapper.find("input[name='externalKey']");
-    expect(usernameInput.prop('defaultValue')).toEqual('');
-    expect(externalKeyInput.prop('defaultValue')).toEqual('');
+    expect(usernameInput.prop('defaultValue')).toEqual(undefined);
+    expect(externalKeyInput.prop('defaultValue')).toEqual(undefined);
```
DawoudSheraz, line 88: huh, why are these props changed? username might have been impacted as part of query param update but external key is unchanged.

src/ProgramEnrollments/ProgramInspector/ProgramInspector.test.jsx
```diff
@@ -188,10 +207,35 @@ describe('Program Inspector', () => {
...
+      .spyOn(api, 'getProgramEnrollmentsInspector')
+      .mockImplementation(() => Promise.resolve(programInspectorSuccessResponse));
+
+    getUserMock = jest
+      .spyOn(ssoAndUserApi, 'getUser')
+      .mockImplementation(() => Promise.reject(new Error('Error fetching User Info')));
+
+    wrapper = mount(<ProgramEnrollmentsWrapper />);
+
+    await updateWrapperState(wrapper);
```
DawoudSheraz: How is this different from waitFor or why is waitFor not applicable here (considering other tests are using it)?

src/users/UserPage.jsx
```diff
@@ -11,7 +11,7 @@ import { isEmail, isValidUsername, isValidLMSUserID } from '../utils/index';
 import { getAllUserData } from './data/api';
 import UserSearch from './UserSearch';
 import LearnerInformation from './LearnerInformation';
-import { LEARNER_INFO_TAB, TAB_PATH_MAP } from '../SupportToolsTab/constants';
+import { TAB_PATH_MAP } from '../SupportToolsTab/constants';
 import CancelRetirement from './account-actions/CancelRetirement';
 
 // Supports urls such as /users/?username={username}, /users/?email={email} and /users/?lms_user_id={lms_user_id}
```
DawoudSheraz: this comment needs an update.

src/users/UserPage.jsx
DawoudSheraz: Right. Shouldn't the method be simplified then (remove username and email part as they will no longer be in query params)?

src/users/UserPage.jsx
DawoudSheraz: Hmmm, the comment above the code mentions that the identifier is added in query param if it is not present (which will happen when the user opens the default page). handleFetchSearchResults does end up calling processSearchResult which calls the same pushHistory method. However, just want to confirm if the flow is working as expected.

zawan-ila: The flow is working fine. Here is what I tried
1. Open the Learner info page
2. Enter a search query (username, email, or lms_id) and hit Search
3. Verify that the url is updated to something like `/learner_information?lms_user_id=X`

src/users/UserPage.jsx
zawan-ila: What method are you referring to?

src/ProgramEnrollments/ProgramInspector/ProgramInspector.jsx
zawan-ila: The query can have username/email inside it. We do not want it to show up in the url. Adding that here would add it to the url.

src/users/UserPage.jsx
zawan-ila: Removing it. It seems to be leftover from some earlier refactoring, considering that we don't really have `/users/` in our urls?

DawoudSheraz: 👍🏽

src/ProgramEnrollments/ProgramInspector/ProgramInspector.jsx
DawoudSheraz: Right. What about the other params in the query like external id?

src/users/UserPage.jsx
DawoudSheraz: This same useEffect where you removed the params 
https://github.com/openedx/frontend-app-support-tools/pull/452/files#diff-ddf5e1127e926c8c62d585e55b5038e0981c9618c5ebf6a32d41234a2ffc0020R126

src/ProgramEnrollments/ProgramInspector/ProgramInspector.test.jsx
zawan-ila: Using `waitFor` now. I couldn't seem to get it to work earlier :-)

src/ProgramEnrollments/ProgramInspector/ProgramInspector.test.jsx
zawan-ila: removed this part.

src/ProgramEnrollments/ProgramInspector/ProgramInspector.test.jsx
zawan-ila: These are coming from the query params. Since we changed the [initial "location"](https://github.com/openedx/frontend-app-support-tools/pull/452/commits/fbdbb1cddeabd2da6158d10799e40f0a4aadb5be#diff-a6d910633c7f2f95caffafa16e1277a3e3730cb1a71eb5ef5ba6deabd939220dL24-R27) for the tests, these are affected. externalKey state is also coming from query params.

src/ProgramEnrollments/ProgramInspector/ProgramInspector.test.jsx
zawan-ila: You might have 2 calls to the method. Say once on initial load, and secondly when manually inputing username etc and hitting search later. This happens in a couple tests. I have removed it from where it is not needed.

src/ProgramEnrollments/ProgramInspector/ProgramInspector.jsx
zawan-ila: That also seems like PII, so got rid of that too. Now we only add lms id in the query params.

src/users/UserPage.jsx
zawan-ila: I left it there as it still allows us to search by `?username=ABC` in the query params. It changes the param to `?lms_user_id` afterwards.

DawoudSheraz (approved):

src/ProgramEnrollments/ProgramInspector/ProgramInspector.jsx
```diff
@@ -41,13 +43,13 @@ export default function ProgramInspector() {
       setSsoRecords([]);
       navigate('/programs');
     } else {
-      const newLink = `/programs?edx_user=${
```
Ali-D-Akbar, line 44: Why did we remove `/programs` again?

src/ProgramEnrollments/ProgramInspector/ProgramInspector.jsx
```diff
@@ -66,13 +68,38 @@ export default function ProgramInspector() {
...
         setLearnerProgramEnrollment(response.learner_program_enrollments);
-      });
+        const name = response?.learner_program_enrollments?.user?.username;
+        return name;
+      }).then((name) => getUser(name)).then((res) => {
+        navigate(`?edx_user_id=${res.id}`);
+      })
+        .catch(err => {
+          console.error(err);
+          setError('An error occured while fetching user id');
```
Ali-D-Akbar: spell: occurred

Ali-D-Akbar: Please update in all places.

src/users/UserPage.test.jsx
```diff
@@ -0,0 +1,65 @@
...
+    jest.spyOn(ssoAndUserApi, 'getVerifiedNameHistory').mockImplementation(() => Promise.resolve(verifiedNameHistoryData));
+    jest.spyOn(ssoAndUserApi, 'getEnrollments').mockImplementation(() => Promise.resolve(enrollmentsData));
+    jest.spyOn(ssoAndUserApi, 'getOnboardingStatus').mockImplementation(() => Promise.resolve(onboardingStatusData));
+    jest.spyOn(ssoAndUserApi, 'getSsoRecords').mockImplementation(() => Promise.resolve(ssoRecordsData));
+    jest.spyOn(ssoAndUserApi, 'getLicense').mockImplementation(() => Promise.resolve(licensesData));
+    jest.spyOn(ssoAndUserApi, 'getEntitlements').mockImplementation(() => Promise.resolve(entitlementsData));
+    jest.spyOn(ssoAndUserApi, 'getEnrollments').mockImplementation(() => Promise.resolve(enrollmentsData));
+    jest.spyOn(ssoAndUserApi, 'getEnterpriseCustomerUsers').mockImplementation(() => Promise.resolve(enterpriseCustomerUsersData));
+
+    jest.clearAllMocks();
```
Ali-D-Akbar, line 48: Shouldn't we have this in afterEach?

src/users/UserPage.test.jsx
```diff
@@ -0,0 +1,65 @@
...
+
+  beforeEach(() => {
+    jest.spyOn(ssoAndUserApi, 'getAllUserData').mockImplementation(() => Promise.resolve({ user: UserSummaryData.userData, errors: [] }));
+    jest.spyOn(ssoAndUserApi, 'getVerifiedNameHistory').mockImplementation(() => Promise.resolve(verifiedNameHistoryData));
+    jest.spyOn(ssoAndUserApi, 'getEnrollments').mockImplementation(() => Promise.resolve(enrollmentsData));
+    jest.spyOn(ssoAndUserApi, 'getOnboardingStatus').mockImplementation(() => Promise.resolve(onboardingStatusData));
+    jest.spyOn(ssoAndUserApi, 'getSsoRecords').mockImplementation(() => Promise.resolve(ssoRecordsData));
+    jest.spyOn(ssoAndUserApi, 'getLicense').mockImplementation(() => Promise.resolve(licensesData));
+    jest.spyOn(ssoAndUserApi, 'getEntitlements').mockImplementation(() => Promise.resolve(entitlementsData));
+    jest.spyOn(ssoAndUserApi, 'getEnrollments').mockImplementation(() => Promise.resolve(enrollmentsData));
```
Ali-D-Akbar: We got a duplicate line here.

src/users/UserPage.test.jsx
```diff
@@ -0,0 +1,65 @@
...
+    jest.spyOn(ssoAndUserApi, 'getEnterpriseCustomerUsers').mockImplementation(() => Promise.resolve(enterpriseCustomerUsersData));
+
+    jest.clearAllMocks();
+  });
+
+  it('default render', async () => {
+    wrapper = mount(<UserPageWrapper />);
+    wrapper.find(
+      "input[name='userIdentifier']",
+    ).instance().value = 'ANonyMouse';
```
Ali-D-Akbar: This may not trigger input change if I'm not wrong... 🤔

src/users/UserPage.test.jsx
```diff
@@ -0,0 +1,65 @@
...
+    wrapper.find('.btn.btn-primary').simulate('click');
+
+    await waitFor(() => {
+      expect(mockedNavigator).toHaveBeenCalledWith(
+        `/learner_information/?lms_user_id=${UserSummaryData.userData.id}`,
+      );
```
Ali-D-Akbar, line 61: Can we add some assertions to check if the API was called?

Ali-D-Akbar: I believe this should call `getAllUserData` if I'm not wrong 😵‍💫

Ali-D-Akbar (commented):
Bunch of comments here n there. Looks good otherwise.

src/ProgramEnrollments/ProgramInspector/ProgramInspector.jsx
zawan-ila: Earlier, we would navigate to this url. Now, we are only using it to store the "query" (backend api call) that we should make. I have refactored it a little. Hopefully it's clearer now. I also removed the `ClickEvent` state, as it wasn't being used anywhere.

src/ProgramEnrollments/ProgramInspector/ProgramInspector.jsx
zawan-ila: Done.

src/users/UserPage.test.jsx
zawan-ila: I feel like it is fine in both `afterEach` and `beforeEach`. Most tests add it in `beforeEach` too.

src/users/UserPage.test.jsx
zawan-ila: Which one?

src/users/UserPage.test.jsx
zawan-ila: Probably, yes. But we don't need a change event anyways.

zawan-ila: Additionally, when I tried a `simulate('onchange'...)`, the value of the input wasn't changing for some reason, and hence the test was failing. Thus, I took this approach.

src/users/UserPage.test.jsx
zawan-ila: Done

src/users/UserPage.test.jsx
Ali-D-Akbar: This line is written twice.

`jest.spyOn(ssoAndUserApi, 'getEnrollments').mockImplementation(() => Promise.resolve(enrollmentsData));`

zawan-ila: removed

Ali-D-Akbar (approved):

hamza-56 (approved):

PR #454, opened 2025-04-08
fix: some issues with program inspector

zawan-ila:
This PR resolves some issues (introduced in [452](https://github.com/openedx/frontend-app-support-tools/pull/452)) with the ProgramInspector :-

1. If a network request fails for a search, running the same search won't retry the request, as the query state has not changed. We change the query from a string type to object type to circumvent this. This is also in line with the earlier behavior (from before [452](https://github.com/openedx/frontend-app-support-tools/pull/452))

2. Remove existing search params whenever a new search is made. Not doing this causes older params to linger for a while until the new search updates the params (asynchronously). Depending on the network, this might take some time.

3. Fix the error message when the getProgramEnrollments call "fails" (e.g when searching for a non-existent user).

### DIFF
--- a/src/ProgramEnrollments/ProgramInspector/ProgramInspector.test.jsx
+++ b/src/ProgramEnrollments/ProgramInspector/ProgramInspector.test.jsx
@@ -228,7 +228,7 @@ describe('Program Inspector', () => {
     await waitFor(() => {
       wrapper.update();
       expect(wrapper.find('Alert').at(0).text()).toEqual('An error occurred while fetching user id');
-      expect(mockedNavigator).toHaveBeenCalledTimes(2);
+      expect(mockedNavigator).toHaveBeenCalledTimes(3);
     });
   });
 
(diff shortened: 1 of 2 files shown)


Ali-D-Akbar (approved):

AfaqShuaib09 (approved):

src/ProgramEnrollments/ProgramInspector/ProgramInspector.test.jsx
```diff
@@ -228,7 +228,7 @@ describe('Program Inspector', () => {
     await waitFor(() => {
       wrapper.update();
       expect(wrapper.find('Alert').at(0).text()).toEqual('An error occurred while fetching user id');
-      expect(mockedNavigator).toHaveBeenCalledTimes(2);
+      expect(mockedNavigator).toHaveBeenCalledTimes(3);
```
DawoudSheraz, line 231: Why is this increasing?

DawoudSheraz (approved):

PR #457, opened 2025-05-14
fix: Resolve email encoding for the payload

hassan-raza-1:
**Description:**
This PR addresses an issue with email encoding when sending data to the API. Previously, email addresses containing special characters—such as the + symbol—were being misinterpreted due to incorrect string handling in the request payload. This fix ensures that email addresses are properly encoded before being sent, preserving their integrity during API communication.

**Support ticket:**
https://2u-internal.atlassian.net/browse/INF-1888

**Testing instruction:**
Submit a password reset request via the Support Tool using an email address that contains a '+' character, and verify that the reset password API call completes successfully.

### DIFF
--- a/src/users/data/api.js
+++ b/src/users/data/api.js
@@ -542,7 +542,15 @@ export async function postTogglePasswordStatus(user, comment) {
 
 export async function postResetPassword(email) {
   try {
-    const { data } = await getAuthenticatedHttpClient().post(AppUrls.getResetPasswordUrl(), `email_from_support_tools=${encodeURIComponent(email)}`);
+    const { data } = await getAuthenticatedHttpClient().post(
+      AppUrls.getResetPasswordUrl(),
+      { email_from_support_tools: email },
+      {
+        headers: {
+          'Content-Type': 'application/x-www-form-urlencoded',
+        },
+      },
+    );
     return data;
   } catch (error) {
     return {
--- a/src/users/data/api.js
+++ b/src/users/data/api.js
@@ -542,7 +542,7 @@ export async function postTogglePasswordStatus(user, comment) {
 
 export async function postResetPassword(email) {
   try {
-    const { data } = await getAuthenticatedHttpClient().post(AppUrls.getResetPasswordUrl(), `email_from_support_tools=${email}`);
+    const { data } = await getAuthenticatedHttpClient().post(AppUrls.getResetPasswordUrl(), `email_from_support_tools=${encodeURIComponent(email)}`);
     return data;
   } catch (error) {
     return {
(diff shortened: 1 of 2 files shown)


hassan-raza-1:
@DawoudSheraz Can you please review this PR

DawoudSheraz (approved):

src/users/data/api.test.js
```diff
@@ -553,7 +553,7 @@ describe('API', () => {
 
     it('Reset Password Response', async () => {
       const expectedResponse = {};
-      mockAdapter.onPost(resetPasswordApiUrl, `email_from_support_tools=${testEmail}`).reply(200, expectedResponse);
+      mockAdapter.onPost(resetPasswordApiUrl, `email_from_support_tools=${encodeURIComponent(testEmail)}`).reply(200, expectedResponse);
```
DawoudSheraz: nit: can you use test.each with different emails (normal characters, normal + special characters) to ensure this always works as expected.

src/users/data/api.js
```diff
@@ -542,7 +542,7 @@ export async function postTogglePasswordStatus(user, comment) {
 
 export async function postResetPassword(email) {
   try {
-    const { data } = await getAuthenticatedHttpClient().post(AppUrls.getResetPasswordUrl(), `email_from_support_tools=${email}`);
+    const { data } = await getAuthenticatedHttpClient().post(AppUrls.getResetPasswordUrl(), `email_from_support_tools=${encodeURIComponent(email)}`);
```
timmc-edx: Instead of passing a string and working with encodings directly, would it work to pass it as a dictionary? I can't tell just from a quick glance at the docs, but I suspect Axios will do the right thing here.

```suggestion
    const { data } = await getAuthenticatedHttpClient().post(AppUrls.getResetPasswordUrl(), {email_from_support_tools: email});
```

hassan-raza-1: You're right — sending the data as a JSON object is generally more standard with Axios, and it would work well under most circumstances. However, in our case, the password_change_request_handler API is a @require_POST Django view that reads data from request.POST. This means it's expecting application/x-www-form-urlencoded data (i.e., a URL-encoded string), not a JSON body.

Changing the frontend to send a JSON object would require corresponding updates on the backend to parse request.body as JSON instead of relying on request.POST. Since this API is used in multiple places, making that change would require updating and testing each usage to ensure we don't break any existing flows.

timmc-edx: Ah yeah, it probably defaults to sending JSON, doesn't it -- so you'd have to tell it to use `application/x-www-form-urlencoded`. https://axios-http.com/docs/post_example gives some examples of that.

timmc-edx: (In any case I think it's fine to do what you've already done here -- just a thought in case you want it to do all the encoding for you.)

hassan-raza-1: I think we should use Axios object syntax with URL-encoded data. It was a  great suggestion.

DawoudSheraz (approved):

zawan-ila (approved):

